PR #299, opened 2014-07-04
pdf generation scripts for getting_started, mix_otp and meta.

liyu1981:
As I found that the files in
https://groups.google.com/forum/#!topic/elixir-lang-core/aRzIbxswsLQ
are gone. I come up with my own scripts to generate nice pdf version
for elixir getting_started documents.

The generated results can be viewed directly as
https://github.com/liyu1981/elixir-lang.github.com/blob/master/_pdf/getting_started.pdf?raw=true
https://github.com/liyu1981/elixir-lang.github.com/blob/master/_pdf/meta.pdf?raw=true
https://github.com/liyu1981/elixir-lang.github.com/blob/master/_pdf/mix_otp.pdf?raw=true


eksperimental:
so it's been two years and two weeks exactly since this has been opened and commented on for the last time.
What do we want to do with this?

another similar projects are, in addition to alco's project is https://github.com/jfreeze/elixir-book

/cc @jfreeze

tvon:
I spent some time on this subject (a PDF of the guide, not related to this PR).  Would the maintainers here be open to PRs that change how the site is managed (e.g., replacing Jekyll with something else)?  I do not have a specific alternative in mind, but it seems like the ideal solution is to have the source material in a state where the tools used to build the site can also generate other formats (instead of maintaining some in-project scripts).  I do not know what tool is ideal for this at the moment, but it would be enough work that I would not want to get too far into it if the idea is a non-starter.

whatyouhide:
I think we use Jekyll for ease of use with GH pages?

tvon:
That was my assumption.  A change in how the site is managed would likely mean some additional overhead in the usual updates.  Maybe just running a rake task, but that is more than you have to do now (which is an assumption on my part).

eksperimental:
given we implement something like this. how is the pdf gonna be generated? automatically on every commit? hosted where?

tvon:
The simplest setup is probably to manually generate the guide whenever that content changes, adn to upload the documents to this repository as releases (or whatever repo is appropriate).

tvon:
We're kind of spitballing here, and I did not mean to hijack this PR.  Maybe this is at a "show me the code" point, or would be better discussed on a mailing list?

liyu1981:
Yes, there are already many decent pdf versions out there, like readthedocs version, or the one mentioned in the comments. I should have closed this earlier :) But thanks for all your comments.

PR #341, opened 2014-08-30
Fix sub title to 'do/end blocks'

yui-knk:


whatyouhide:
Do we have any news on this? It's a pretty trivial addition :smile:

josevalim:
I completely missed it!

whatyouhide:
@josevalim that's what we annoying contributors are here for :)

PR #353, opened 2014-09-07
Fixed Chapter 11 Link section, new Logger output made it confusing

asonge:
Fixed Chapter 11 Link section, which was made confusing by the introduction of Logger

The Logger outputs something, which makes the "nothing happens" phrase confusing for people using the Getting Started Guide.


josevalim:
We hve updated this to talk about tasks

PR #388, opened 2014-11-15
update test to avoid race condition, add note explaining change

montague:
I was following the chapter on GenServer and noticed a race condition in the test file that caused intermittent failures. I updated the test code and added a note about why it was necessary.

The other approach would be not to use ```async: true``` in that test example, but I thought it was educational to call attention to the race condition.

Specifically, when running with ```async: true```, about one out of ten times, the final assertion
```
assert KV.Registry.lookup(registry, "shopping") == :error
```
would fail because (I think) the stopped bucket would have been restarted by the previous test prior the assertion above.


josevalim:
Fixed in upcoming 1.2 branch.

PR #403, opened 2015-01-02
Fix html

eksperimental:
Fixes several broken HTML issues, based on tidy-html5 validator,
and other improvements

One issue is now that the html is fixed, 
the titles for the entries in http://elixir-lang.org/blog/
look different,
now they look a bit smaller (16px, vs 24px before)
so If we want bitter titles, I  need to patch the css

### DIFF
--- a/_includes/bottom.html
+++ b/_includes/bottom.html
@@ -4,7 +4,7 @@
       </div>
 
       <div id="copyright">
-        © 2012-2014 <a href="http://plataformatec.com.br/">Plataformatec</a>. All rights reserved.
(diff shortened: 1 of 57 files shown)


eksperimental:
Oh, i forgot to mention,
as for the elixir-guide,
i have moved the Table of contents to a template,
so instead of using 
````  <div class="toc"></div>````

we should use:
`{% include toc.html %}`

josevalim:
Merged, thank you! We shuold definitely increase the title size on the blog page, can you please send a new pull request?

eksperimental:
You are welcome!
and fixed!

PR #406, opened 2015-01-03
fix entry title font size in /blog/

eksperimental:
This fixes the issue with the size of the title mentioned in #403


josevalim:

Issue #399, opened 2014-12-27
8. Modules - a default value - an example


nedzadarek:
An example doesn't show that a default argument (value) is evaluated every time we need it.
```elixir
defmodule DefaultTest do
  def dowork(x \\ IO.puts "hello") do
    x
  end
end
iex> DefaultTest.dowork 123
123
iex> DefaultTest.dowork
hello
:ok
```

I would do something like this:

```elixir
 DefaultTest.dowork 
 DefaultTest.dowork 123
 DefaultTest.dowork 
```

PR #410, opened 2015-01-06
Example of default args being always evaluated (ch. 8, Modules)

whatyouhide:
Exactly what @nedzadarek described in #399 :smile:.

### DIFF
--- a/getting_started/8.markdown
+++ b/getting_started/8.markdown
@@ -185,6 +185,9 @@ end
 ```
 
 ```iex
+iex> DefaultTest.dowork
+hello
+:ok
 iex> DefaultTest.dowork 123
 123
 iex> DefaultTest.dowork


getting_started/8.markdown
```diff
@@ -185,6 +185,9 @@ end
 ```
 
 ```iex
+iex> DefaultTest.dowork
+hello
+:ok
 iex> DefaultTest.dowork 123
 123
 iex> DefaultTest.dowork
```
josevalim: Isn't this what is shown here 3 lines below? :D Maybe we should remove the ones at the end?

josevalim: Ah misunderstood it then. :( I don't see the purpose in invoking the same function twice. I thought it was never invoked. :(

whatyouhide: If you want, I can just change it to

``` elixir
DefaultTest.dowork
DefaultTest.dowork 123
```

or

``` elixir
DefaultTest.dowork 123
DefaultTest.dowork
```

but I'm actually for the version in this PR. Both the versions above kind of fail to show that the default argument is evaluated **every** call: the first version shows that the default argument is evaluated the first time but doesn't actually show that it's re-evaluated on later calls; the second version shows that the default argument is evaluated even if that's not the first call to the function, but fails to show that if you call the function again with no arguments it will still be evaluated, again.

I think that not showing the call to `dowork` without arguments at least twice could lead people into thinking that the first call with the default argument *evaluates* it, but then the runtime stores it and successive calls don't evaluate it or similar scenarios. Showing the call twice should avoid arising such doubts.

What do you think?

josevalim:
I see now, perfect. :)

Issue #395, opened 2014-12-13
Getting started docs do not introduce the pipe operator


IdahoEv:
Working through as a beginner (and using it as a reference), I notice that the |> operator is never actually explained in "Getting Started".  It first appears  in example code in 10.1 "Eager vs. Lazy", where it is used without explanation or introduction.

I would like to insert a new section to introduce the pipe operator.  If you're okay with this proposal I'll get writing and issue a PR. 

My instinct would be to put the new section between 5 and 6, after other basic language structures but before data types. It would probably also be okay later on. It would have two subsections:

### subsection 1: pipe operator and method chaining
introduce the operator and explain how it improves readability over f4(f3(f2(f1(arg))))

### subsection 2: chainable methods
Introduce this stylistic / semantic tradeoff:

* Erl-style functions that return { :ok | :error, value }, which are useful for handling error cases with pattern matching
* Simple return functions that just return value, which are easier to use with |>.    

Then explain how the Elixir stdlib sometimes introduces both variants for this reason (E.g. Dict.fetch/2 and Dict.fetch!/2)

whatyouhide:
I think this can be closed since #411 has been merged :).

PR #411, opened 2015-01-07
Brief description of the pipe operator in chapter 10

whatyouhide:
I added a brief description of the pipe operator (`|>`) in chapter 10, when it is first used. This has been suggested by @josevalim in #395.

### DIFF
--- a/getting_started/10.markdown
+++ b/getting_started/10.markdown
@@ -54,18 +54,28 @@ iex> 1..100_000 |> Enum.map(&(&1 * 3)) |> Enum.filter(odd?) |> Enum.sum
 
 The example above has a pipeline of operations. We start with a range and then multiply each element in the range by 3. This first operation will now create and return a list with `100_000` items. Then we keep all odd elements from the list, generating a new list, now with `50_000` items, and then we sum all entries.
 
-As an alternative, Elixir provides [the `Stream` module](/docs/stable/elixir/Stream.html) which supports lazy operations:
+### 10.2.1 The pipe operator
+
+The `|>` symbol used in the snippet above is the **pipe operator**: it simply takes the output from the expression on its left side and passes it as the input to the function call on its right side. It's similar to the Unix `|` operator.  Its purpose is to highlight the flow of data being transformed by a series of functions. To see how it can make the code cleaner, have a look at the example above rewritten without using the `|>` operator:
+
+```iex
+iex> Enum.sum(Enum.filter(Enum.map(1..100_000, &(&1 * 3)), odd?))
+7500000000
+```
+
+Find more about the pipe operator [by reading its documentation](http://elixir-lang.org/docs/stable/elixir/Kernel.html#|>/2).
+
+## 10.3 Streams
+
+As an alternative to `Enum`, Elixir provides [the `Stream` module](/docs/stable/elixir/Stream.html) which supports lazy operations:
 
 ```iex
 iex> 1..100_000 |> Stream.map(&(&1 * 3)) |> Stream.filter(odd?) |> Enum.sum
 7500000000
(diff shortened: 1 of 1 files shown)


getting_started/10.markdown
```diff
@@ -54,7 +54,16 @@ iex> 1..100_000 |> Enum.map(&(&1 * 3)) |> Enum.filter(odd?) |> Enum.sum
...
+### The pipe operator
+
+The `|>` symbol used in the snippet above is the **pipe operator**: it simply takes the output from the expression on its left side and passes it as the input to the function call on its right side. It's similar to the Unix `|` operator.  Its purpose is to highlight the flow of data being transformed by a series of functions. To see how it can make the code cleaner, have a look at the example above rewritten without using the `|>` operator:
+
+```iex
+iex> Enum.sum(Enum.filter(Enum.map(1..100_000, &(&1 * 3)), odd?))
+7500000000
+```
+
+As an alternative to `Enum`, Elixir provides [the `Stream` module](/docs/stable/elixir/Stream.html) which supports lazy operations:
```
josevalim: I believe we could move this and the following chapters inside section `10.3`.

whatyouhide: I agree to move from line 66 on to section 10.3, but I think it would be better to leave the "The pipe operator" subsection at the end of section 10.2 since that's where the operator is introduced. I feel like in section 10.3 it would be a little out of context.

What do you think?

getting_started/10.markdown
```diff
@@ -54,7 +54,16 @@ iex> 1..100_000 |> Enum.map(&(&1 * 3)) |> Enum.filter(odd?) |> Enum.sum
...
-As an alternative, Elixir provides [the `Stream` module](/docs/stable/elixir/Stream.html) which supports lazy operations:
+### The pipe operator
+
+The `|>` symbol used in the snippet above is the **pipe operator**: it simply takes the output from the expression on its left side and passes it as the input to the function call on its right side. It's similar to the Unix `|` operator.  Its purpose is to highlight the flow of data being transformed by a series of functions. To see how it can make the code cleaner, have a look at the example above rewritten without using the `|>` operator:
+
+```iex
+iex> Enum.sum(Enum.filter(Enum.map(1..100_000, &(&1 * 3)), odd?))
+7500000000
+```
+
```
josevalim: We could end this section on the pipe operator with something like: "Find more about the pipe operator [by reading its documention]." With a link, of coure. :)

getting_started/10.markdown
```diff
@@ -54,7 +54,16 @@ iex> 1..100_000 |> Enum.map(&(&1 * 3)) |> Enum.filter(odd?) |> Enum.sum
 
 The example above has a pipeline of operations. We start with a range and then multiply each element in the range by 3. This first operation will now create and return a list with `100_000` items. Then we keep all odd elements from the list, generating a new list, now with `50_000` items, and then we sum all entries.
 
-As an alternative, Elixir provides [the `Stream` module](/docs/stable/elixir/Stream.html) which supports lazy operations:
+### The pipe operator
```
josevalim: We need to label this accordingly as `10.2.1 The pipe operator`.

getting_started/10.markdown
josevalim: > I agree to move from line 66 on to section 10.3, but I think it would be better to leave the "The pipe operator" subsection at the end of section 10.2 since that's where the operator is introduced. I feel like in section 10.3 it would be a little out of context.

This is exactly what I had in mind, yes. :)

whatyouhide:
Pushed all the changes we discussed. Let me know if I missed something :).

josevalim:

Issue #31, opened 2012-06-09
Include a chapter about behaviors


josevalim:

whatyouhide:
The chapter about behaviours (and typespecs) has been added in #412, so I think this can be closed! :+1:

josevalim:
:heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

Issue #32, opened 2012-06-09
Include a chapter about typespecs


josevalim:

whatyouhide:
I think this could be closed after #412 has been merged?

josevalim:
:heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #412, opened 2015-01-08
Chapter about typespecs and behaviours

whatyouhide:
I stubbed a first draft of a possible chapter about behaviours. I saw that #31, the issue about such chapter, is quite old (mid 2012!), but it looks like it's never been tackled.

I pushed a "draft" commit, so that you can have a look at how things have came out. I'm actually not 100% sure you still need a chapter about behaviours (or **only** about behaviours, for that matter). I positioned the chapter as the last one (21) in the getting started guide navigation, but didn't actually number it (the file is called `behaviours.markdown` and the sections are numbered as `x.1` and `x.2`).

Let me know what you think of all this :smiley:. Thanks!

### DIFF
--- a/_layouts/getting_started.html
+++ b/_layouts/getting_started.html
@@ -29,7 +29,8 @@ section: getting_started
       <li><a class="spec" href="/getting_started/17.html">try, catch and rescue</a></li>
       <li><a class="spec" href="/getting_started/18.html">Comprehensions</a></li>
       <li><a class="spec" href="/getting_started/19.html">Sigils</a></li>
-      <li><a class="spec" href="/getting_started/20.html">Where to go next</a></li>
+      <li><a class="spec" href="/getting_started/20.html">Typespecs and behaviours</a></li>
+      <li><a class="spec" href="/getting_started/21.html">Where to go next</a></li>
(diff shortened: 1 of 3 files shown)


josevalim:
Excellent, thank you! The main issue about behaviour though is that they require us to talk about typespecs. So a quick introduction to typespecs plus behaviours would likely be a great chapter, so +1 for this addition. We can place it after the sigils one.

whatyouhide:
I agree 100% on combining typespecs and behaviours in a single chapters. I started with behaviours only since #31 is only about behaviours and #32 is about typespecs, so I assumed they should've been separate chapters. I'll try to make a draft of the "Typespecs and behaviours" chapter as soon as I can.

whatyouhide:
I updated the chapter making it a chapter about **typespecs** and **behaviours**. I placed the chapter after the "Sigils" chapter, making it chapter 20. I also updated the chapter numbers and the navbar so that the "Where to go next?" chapter is now chapter 21.

I'm sure a lot of things can be written in a better way, so just let me know what you think for now :smiley:. Thanks!

josevalim:

PR #417, opened 2015-01-13
simplify css border images on widgets + remove border from book covers

eksperimental:
Remove the border from the images on the right-hand side-bar (some book covers end up with a double border), and only add it through a "border" class  to the ones that needed.


josevalim:

PR #428, opened 2015-01-18
Rename chapter 16 and fix a bunch of errors in that chapter

whatyouhide:
Chapter 16 was called just "IO" but explained both how to do IO in Elixir as well as how to interact with files and the filesystem (the `File` and `Path` modules). The new title - "IO and interacting with the filesystem" - reflects exactly that.

I also fixed a bunch of grammar/spelling errors throughout the file and rephrased some sentences here and there. There are a few comments in the diff justifying the changes :smile:.


josevalim:
I don't like the new title, it is too long. :P Maybe we can say `IO and the filesystem` or just leave it as `IO`. Thoughts?

Thanks again!

whatyouhide:
I'm reluctant to leaving it as just "IO" since, well, it deals with the filesystem too :dancers: I've `push --force`ed the "IO and the file system" chapter title, let me know what you think :smiley:

josevalim:

PR #429, opened 2015-01-19
Very small polishes in chapter 13

whatyouhide:
- Fixed some minor grammar errors
- Rephrased a couple of sentences
- Added a note at the very end of the chapter (about module nesting)

As usual, most changes are justified in the diff's comments.

### DIFF
--- a/getting_started/13.markdown
+++ b/getting_started/13.markdown
@@ -20,7 +20,7 @@ defmodule Math do
 end
 ```
 
-From now on, any reference to `List` will automatically expand to `Math.List`. In case one wants to access the original `List`, it can be done by accessing the module via `Elixir`:
+From now on, any reference to `List` will automatically expand to `Math.List`. In case one wants to access the original `List`, it can be done by prefixing the module name with `Elixir.`:
 
 ```elixir
 List.flatten             #=> uses Math.List.flatten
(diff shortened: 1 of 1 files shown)


getting_started/13.markdown
```diff
@@ -176,4 +181,8 @@ defmodule Elixir.Foo do
 end
 ```
 
-As we will see in later chapters, aliases also play a crucial role in macros, to guarantee they are hygienic. With this we are almost finishing our tour about Elixir modules, the last topic to cover is module attributes.
+**Note**: in Elixir, you don't have to define the `Foo` module before being able to define the `Foo.Bar` module, as the language translates all module names to atoms anyway. You can define arbitrarly-nested modules without defining any module in the chaing (e.g., `Foo.Bar.Baz` without defining `Foo` or `Foo.Bar` first).
```
josevalim: chaing -> chain

getting_started/13.markdown
```diff
@@ -164,7 +164,12 @@ defmodule Foo do
 end
 ```
 
-The example above will define two modules `Foo` and `Foo.Bar`. The second can be accessed as `Bar` inside `Foo` as long as they are in the same lexical scope. If later the developer decides to move `Bar` to another file, it will need to be referenced by its full name (`Foo.Bar`) or an alias needs to be set using the `alias` directive discussed above.
+The example above will define two modules: `Foo` and `Foo.Bar`. The second can be accessed as `Bar` inside `Foo` as long as they are in the same lexical scope. If later the `Bar` module is moved outside the `Foo` module definition, it will need to be referenced by its full name (`Foo.Bar`) or an alias will need to be set using the `alias` directive discussed above. The `Bar` module definition will change too:
+
+```elixir
+defmodule Foo.Bar do
+end
```
josevalim: I think you are missing an alias in this code sample:

```elixir
alias Foo.Bar, as: Bar
defmodule Foo.Bar do
end
```

whatyouhide: @josevalim I was actually only showing that if you move `Bar` outside `Foo` but still want `Bar` to be "nested" inside `Foo`, you have to define `Bar` with its fully qualified name, `Foo.Bar`. I'm gonna rephrase and push a new version in a few minutes.

whatyouhide:
@josevalim Pushed a new commit where I fixed the typo and rephrased the example we discussed. Let me know what you think!

getting_started/13.markdown
```diff
@@ -164,7 +164,16 @@ defmodule Foo do
 end
 ```
 
-The example above will define two modules `Foo` and `Foo.Bar`. The second can be accessed as `Bar` inside `Foo` as long as they are in the same lexical scope. If later the developer decides to move `Bar` to another file, it will need to be referenced by its full name (`Foo.Bar`) or an alias needs to be set using the `alias` directive discussed above.
+The example above will define two modules: `Foo` and `Foo.Bar`. The second can be accessed as `Bar` inside `Foo` as long as they are in the same lexical scope. If later the `Bar` module is moved outside the `Foo` module definition, it will need to be referenced by its full name (`Foo.Bar`) or an alias will need to be set using the `alias` directive discussed above. The `Bar` module definition will change too. This code is equivalent to the example above:
```
josevalim: So I would start a new paragraph on "If later the Bar module...".

whatyouhide: Got it. Pushed a new version :smile:

PR #433, opened 2015-01-30
Learning resources updated

eksperimental:
* New "Metaprogramming Elixir" book added
* Link's titles now contain: Book Title: Subtitle: Author
* Update Programming Elixir link to avoid redirect
* Minor Fix install.html, as it wasn't validating

### DIFF
--- a/_includes/learning-resources.html
+++ b/_includes/learning-resources.html
@@ -1,10 +1,15 @@
 <div class="widget">
   <h3 class="widget-title">Learning resources</h3>
   <ul>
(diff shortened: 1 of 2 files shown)


eksperimental:
my only question is whether it is ok to load images from an https server 
(both pragmatic programming book's images)
The are redirected anyway, but I'm not sure if it's advisable to do it (I know the issue are for the other way around scenarios - trying to load http from https), just let me know so I update the code if need

josevalim:
It is probably ok. We shouldn't add the meta-programming cover just yet. I had other books to add and I promised them that I would only do it when I revamp the learning resources page. So, to be fair to everyone, it needs to wait too. :)

I will give the learning section higher priority. Can you please remove it for now?

eksperimental:
sure Jose, I will remove just that title

eksperimental:
we are good to merge!

eksperimental:
the image is also provided with http (no redirect), so no need to change anything. it's only with the page that the redirection is happening thus the image with https was being loaded

PR #434, opened 2015-01-30
Added examples of non-last default arguments to chapter 8

whatyouhide:
Default arguments need not be the last argument in a function signature; I added an example of that at the end of chapter 8.

### DIFF
--- a/getting_started/8.markdown
+++ b/getting_started/8.markdown
@@ -42,7 +42,7 @@ defmodule Math do
 end
 ```
 
-This file can be compiled using `elixirc`:
+This file can be compiled using `elixirc`, the Elixir compiler:
 
     $ elixirc math.ex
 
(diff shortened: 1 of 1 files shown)


getting_started/8.markdown
```diff
@@ -53,12 +53,6 @@ iex> Math.sum(1, 2)
 3
 ```
 
-Elixir projects are usually organized into three directories:
```
josevalim: Why is this going away?

getting_started/8.markdown
```diff
@@ -42,7 +42,7 @@ defmodule Math do
 end
 ```
 
-This file can be compiled using `elixirc`:
+This file can be compiled using `elixirc`, the Elixir compiler:
```
josevalim: Strinctly speaking, it is not the Elixir compiler, just an interface to it.

josevalim:
Thank you but I think I disagree with this change. The getting started guide is a getting started guide, not a complete reference and talking about optional arguments in other places do not add much. Plus, it should be used only in strict cases so we can leave it for the developers to find out later. :)

whatyouhide:
@josevalim I was afraid of that :) thanks anyways!

PR #435, opened 2015-02-01
Update example in Sigils chapter to corretly return negative integer

Takemikazuchi:
I was going through the tutorial and noticed this error.


josevalim:

PR #438, opened 2015-02-02
Add more details to unquoting in meta/chapter-1

whatyouhide:
I added an additional code sample showing what happens when you don't `unquote` a value but you want to inject that value in a quoted expression.


josevalim:

Issue #442, opened 2015-02-04
Getting Started Guide: rename files by "name"


eksperimental:
There is one crucial issue with the Getting Started Guide: File names use numbers, not names. This will cause the site to eventually break when pages are removed, or new pages are added, as it's currently happening, and [planned](#405).

In order to to this we need to device a way to store the page order for each section of the guide. I've been looking at a few solutions, but first I need to hear opinions.

whatyouhide:
:+1: all the way! Problem is we don't want to break all existing links to the current chapters either, but I can't think of a way to "redirect" using Jekyll.

eksperimental:
there is a plugin for that

eksperimental:
https://github.com/jekyll/jekyll-redirect-from

Just add this code to the top of the page (YAML frontmatter)
```
redirect_from:
  - /getting_started/mix_otp/2.html
```

whatyouhide:
@eksperimental Yeah I had a look at that and at https://github.com/nquinlan/jekyll-pageless-redirects as soon as you commented :). Nice stuff, would definitely help with this issue.

eksperimental:
@whatyouhide your option is even better for this job

josevalim:
Yeah, great point! let's migrate from numbers!

-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Lead Developer

whatyouhide:
@eksperimental is right though, we have to think of a way to store the order of the chapters in the guides.

eksperimental:
i found two possible solutions: non of them tested,
http://stackoverflow.com/a/16625534/1660683
http://stackoverflow.com/a/25513956/1660683

and we already have the value
`guide: `
which is the order (same number as the file name)

eksperimental:
i think the examples above don't suit our needs,
i have create the structure:
https://gist.github.com/eksperimental/4d7b1c392f6d401bd4c5

and it should be saved in: `_data/getting_started.yml`

I will see if I can come up with something good

josevalim:
The index in the sidebar is the order, no?



-- 


*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Lead Developer*

eksperimental:
you talking about this file?
https://gist.githubusercontent.com/eksperimental/4d7b1c392f6d401bd4c5/raw/c987ee755460c6c546d721c89c64ddc0bd5a4e93/getting_started

[...]

PR #443, opened 2015-02-04
Getting Started: get rid of numeric pages

eksperimental:
Solves #442

All done (or almost).
the tree structure of the guides is all configurable in one file `_data/getting_started.yml`
from there the side bar is created, and the navigation for every page.
I have installed the [pageless-redirect](https://github.com/jekyll/jekyll-redirect-from plugin), it's a patched version from the PR in their github account.

you can see the redirection mapping in `_redirects.yml`
one improvement i made, is that `/getting_started/` redirects for the first chapter of the guide, and same for `/getting_started/meta/`, and `/getting_started/mix_otp`

I have put the site through a validator, and it's all good.
I checked all the links... there are a few dead links to fix (that's for another PR), but nothing introduced in these changes.
I uploaded it to my page, but it doesn't work since it runs on a subfolder.

please fetch this PR, and run `jekyll serve`

and thing we are all good.
The only repeated information is this

```
title: 13 alias, require and import
guide: 13
```

it's easier to load directly from the page.
We can still get rid of `guide:`,

And if a new page is introduced, we just need to replace 13, with 14 in the title and the headings for example `## 13.1 alias`

TODO: update all the links to avoid the redirection.


eksperimental:
@whatyouhide if you want to download this pull request, do this:
```
git fetch upstream pull/443/head:page_names
git checkout page_names
```

whatyouhide:
@eksperimental did you just delete all the files in the guides in 4d6d21f?

eksperimental:
please, refetch, i just  pushed --force
5 minutes ago.
yeah. something was wrong with the first commit, files were deleted

whatyouhide:
Looks nice. So there's no way to get rid of the `title:` and `guide:` metadata? Some way to infer them from `_data/getting_started.yml`? I always hated having to manually change chapter (and section) numbers.

I'm not familiar with Jekyll at all, that's why I'm just asking questions instead trying to solve this myself :). I don't know how Jekyll handles "helpers". If we could write our own helpers, we could avoid doing `## Section title` and write something like `<%= section('Section title') %>` so that section are automatically numbered.

eksperimental:
as I mentioned before, we can get rid of `guide`,
We could leave guide at the `top`, ans reuse it along the artile, but i think it's better to have 
`## 13.1 alias`
than to have
`## {{guide}}.1 alias``
when you are in writing mode, i found it to be distracting all that.

same here. i don't know much about jekyll, just what i grasped from working on this website.
a solution like the one you proposed would be good.

whatyouhide:
I prefer `## 13.1 foo` over `## {{guide}}.1 foo` too if we can't avoid writing the section number `.1` :). I've always used [Middleman](https://middlemanapp.com) as a static site generator and always found it much more lean and intuitive (also more Rails-like) than Jekyll.

I'd be happy to invest a couple of days into migrating the site to Middleman but I suspect @josevalim and the rest of the team wouldn't be happy with such major changes.

josevalim:
Yes, let's migrate to `getting-started` and `mix-otp` too! Great, great work!

eksperimental:
thank you Jose.
I think i'm done, for the first part
All links have been replaced, still the old links redirecting,
I have disabled the redirection temporarily (for testing purposes), and crawled the site, and no dead links, so we are good within the site even without redirection.
every single occurrence of "getting_started"  has been replaced with "getting-started", including filenames and folders
no dead links in the website, including external links (except for this one in docs: http://elixir-lang.org/docs/stable/elixir/Version.Requirement.html#t:t/0  that @josevalim  is taking care of).

It has been thoroughly tested, but since many changes and it has been automated, error may exist.
so please test.

All pages validate except for one little thing I will report later on (the attribute "X-UA-Compatible" not being recognized by the validator)

have a look guys and tell me what you think

eksperimental:
There is one thing left, and is this: https://github.com/elixir-lang/elixir-lang.github.com/issues/442#issuecomment-72990075

I think it is important not to have numbers in the links, but unless we use a custom jekyll plugin, or hack the javascript that creates the TOC, we have to drop the numbers, at least in the subheaders..
if you are ok with that. Numbers look good before every heading,

josevalim:
:heart:

josevalim:
Let's proceed with the TOC changes too as you find appropriate!

eksperimental:
alright, i will do those changes.

one problem: redirectionis not working
http://elixir-lang.org/getting_started/1.html

eksperimental:
bad news!
github doesn' t support custom plugins. I totally overlooked this aspect
:-(

josevalim:
No problem. Let's remove the custom plugin and use the redirect_from we mentioned earlier (which I have used in the past and it does work!).

eksperimental:
let me try something easier. i will post a PR in 10 minutes.
and you tell me which one you prefer

eksperimental:
yeah.. i don' t like much my solutions (I won' t even PR), https://github.com/eksperimental/elixir-lang.github.com/tree/page_names
it was createing a `/getting_started/` folder and holding all the redirect html files

I will proceed with `redirect_from`

josevalim:
That's basically what redirect_from will do. Btw, can you send a pull request in two commits? One that removes the current plugin and another one that adds redirect_from? This way we can easily remove redirect_from changes in the future. :)

PR #445, opened 2015-02-06
Use "redirect_from" plugin

eksperimental:
this removes `jekyll-pageless-redirects` and add `redirect-from` solving https://github.com/elixir-lang/elixir-lang.github.com/pull/443#issuecomment-73162091


josevalim:

Issue #386, opened 2014-11-13
Create a "directory listing" page for http://elixir-lang.org/docs/stable/ and http://elixir-lang.org/docs/master/ 


eksperimental:
I find it a bit unintuitive to have an ERROR 404 page when visiting: http://elixir-lang.org/docs/stable/ and http://elixir-lang.org/docs/master/

I think it should be helpful to list all the projects under that folder.

eksperimental:
I have submitted a pull request to /docs

https://github.com/elixir-lang/docs/pull/5

let me know whether it works or not, as i wasn't able to set up a gh-page in my respository.

josevalim:
:heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #446, opened 2015-02-06
index-redirect.html added

eksperimental:
This addition solves the issue of visiting a the folder, with no index.html
http://elixir-lang.org/getting_started/mix_otp/ well, now http://elixir-lang.org/getting-started/mix-otp/
where should it go.

This include file read _data/getting-started or whatever file you want, look in the document structure for the current folder, and gives redirects you to the first article.

I created it from scratch. And I think it can be reused in other jekyll projects

### DIFF
--- a/_includes/index-redirect.html
+++ b/_includes/index-redirect.html
@@ -1,0 +1,68 @@
+{% comment%}
+  Description:
+    This include file redirects you from the folder/ to folder/first_page_in_your_series.html
+
+  Usage:
+    Include this file using {% include redirect.html %},
+    You can assign in the Frontmatter two variables:
+    
+    `data_file: getting-started`
+      Specify the name of the file under: _data/
+      The slug for the first article listed in the current folder, will be picked.
+      The redirect_url will be create using: `current_directory + slug + .html`
+
+    `redirect_url: /some_dir/some_page#internal_link`
+      or you can directly set the address.
+
+    Additionally you can set your own custom message, setting the variable `custom_content`.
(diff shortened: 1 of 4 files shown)


josevalim:
Given the previous addresses never worked. I don't think we need to worry about redirect. We can just provide the index page or we can just not add those files at all (since we never had them).

eksperimental:
Well, it's the same thing as i mentioned a while ago with https://github.com/elixir-lang/elixir-lang.github.com/issues/386

There are two valid points:
1. it's much easier to remember when you want to link or visit, and you can write type almost off the top of your head.
2. when you want to refer to the guide and link to it, let's say, to Meta Programming: 
http://elixir-lang.org/getting-started/meta/ will always link to the first chapter,
whereas http://elixir-lang.org/getting-started/meta/quote-and-unquote.html may not be the beginning of the guide in the future.

Also an index is not a bad idea, actually, it's a really good one, and it can be populated from the _data/getting-started file
I think having something is better than not having it at all (regardless we hadn't  it before)

josevalim:

eksperimental:
I was writting, "Indeed, an index page is a much better solution for this"
before i hit "comment" it's been merge ;-?

josevalim:
hahah, we can change it to an index page too. :)

eksperimental:
no worries :+1:  :-D 
btw, whenever you get the change don't forget to have a look at this https://github.com/elixir-lang/docs/pull/7#issuecomment-72585694

PR #448, opened 2015-02-06
Getting started headings

eksperimental:
this removes the numbering from the headings (avoiding internal links to break over time)
and a lot of formatting fiddling in the Getting Started guide

### DIFF
--- a/_includes/bottom.html
+++ b/_includes/bottom.html
@@ -26,7 +26,7 @@
     $(document).ready(function() {
         $('.toc').toc({
           title: '',
-          listType: 'ul',
+          listType: 'ol',
           minimumHeaders: 2,
           headers: 'h2, h3, h4, h5, h6',
           showSpeed: 0,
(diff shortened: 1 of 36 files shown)


josevalim:
Thank you! Does this mean the headings will no longer be followed by their number?

eksperimental:
yes. (I like it better with the numbers, but due to the limitations of being hosted in github pages)..
It will be good if we could do it in javascript.

whatyouhide:
I'm not very :+1: on not having numbers before the section header. Doing in JS should be trivial, I'd be happy to do it by tonight if this gets merged soonish :).

josevalim:
Awesome, I am merging this and then we can add the JS stuff.

eksperimental:
sounds great, I realized "install" and "crash-course" needs to have the numbers removed.
thanks a lot @whatyouhide for joining in.

PR #456, opened 2015-02-11
Improve jekyll toc, octicons added

eksperimental:
I have been working on this for the last days, it is not much, but i think it's a refined behaviour that we can translate to the elixir docs. 

It copied the idea from Github, when you can link to a heading, and i have added a back-to-top icon, on the right hand side.

I have installed the [octicon package](https://octicons.github.com/), which is a free font realized by github, with the icons they use in the site, but vectored into a font (so you can scale it and the quality will still be the same)

you can click on the link icon that appears when you mouseover,
or you can click directly on the heading text (i chose not to change the cursos to a pointer in this situation to not to create too much distraction and be as unobstrusive as possible)
if you got the the right hand side, you can click on the back-to-top icon ( ^ ). and that takes you wherever you set it to, in this case, the top of the TOC. when you go back it goes with a transition effect, 

tell me guys what you think,
I hope you like it and find it useful.

here's a screencast
new-toc2_optimized_o3_32


josevalim:
This looks great! I am just worried about octicons. It seems to be a lot to add (fonts + css files) for very little benefit. Maybe we should just hand roll or extract the bits we care?

eksperimental:
Yes i thought of that,
do you have an idea how to extract these chars,
I have already posted a question in the octicon package asking for the building scripts.

josevalim:
Maybe we can extract them from the svg file? Not sure if svg is supported on all recent browsers though.

eksperimental:
I think we are sorted, i found this app online
https://icomoon.io/app/

It creates a set of fonts with whatever icons from any fonts you want, and it creates the fonts in all the same formats the octicons package support (svg, eot, ttf, woff)
all four fonts = 6Kb

josevalim:
Beautiful! :+1: Just remember to keep octicons copyright notices! :)

eksperimental:
jose, we are good to go

josevalim:
Thank you! I have merged it and Im noticing some weird behaviour:

1. The icons are always showing up!

screen shot 2015-02-15 at 12 39 16

2. There is some weird flickering when click on the title due to the animation. Can we remove the animation? I don't think it helps much and there is always the chance of it backfiring. :)

eksperimental:
I think if fixed the issue with the animation, but there is a change that it is still happening, i just set the duration of it to zero (but i fixed what i think it was),
and the other thing was a last minute change  in the css.
see if it works fine
if it's still flickering, i will remove the animation from the code completely.

PR #460, opened 2015-02-12
Fix GenEven test

### LINKED ISSUE #459: GenEvent: Example not working
the example in the GenEvent page, is not working

http://elixir-lang.org/getting-started/mix-otp/genevent.html#registry-events

```elixir
defmodule Forwarder do
  use GenEvent

  def handle_event(event, parent) do
    send parent, event
    {:ok, parent}
  end
end

setup do
  {:ok, manager} = GenEvent.start_link
  {:ok, registry} = KV.Registry.start_link(manager)

  GenEvent.add_mon_handler(manager, Forwarder, self())
  {:ok, registry: registry}
end

test "sends events on create and crash", %{registry: registry} do
  KV.Registry.create(registry, "shopping")
  {:ok, bucket} = KV.Registry.lookup(registry, "shopping")
  assert_receive {:create, "shopping", ^bucket}

  Agent.stop(bucket)
  assert_receive {:exit, "shopping", ^bucket}
end
```

I have tried to fix it, it, but i cannot make `assert_recive` work
the registry seems to be empty.

```elixir
defmodule KV.RegistryTest do
  use ExUnit.Case, async: true

  defmodule Forwarder do
    use GenEvent

    def handle_event(event, parent) do
      send parent, event
      {:ok, parent}
    end
  end

  setup do
    {:ok, manager} = GenEvent.start_link
    {:ok, registry} = KV.Registry.start_link

    GenEvent.add_mon_handler(manager, Forwarder, self())
    {:ok, registry: registry}
  end

  test "sends events on create and crash", %{registry: registry} do
    KV.Registry.create(registry, "shopping")
    {:ok, bucket} = KV.Registry.lookup(registry, "shopping")
    assert_receive {:create, "shopping", ^bucket}

    Agent.stop(bucket)
    assert_receive {:exit, "shopping", ^bucket}
  end
end
```

eksperimental:
* Fix GenEven test (fixes #459 )
* Correction in Supervisor runing `iex` with _run_ option (José, could you check this is correct, otherwise it needs rephrasing as it is not clear)

### DIFF
--- a/getting-started/mix-otp/genevent.markdown
+++ b/getting-started/mix-otp/genevent.markdown
@@ -73,31 +73,31 @@ In order to emit events, we need to change the registry to work with an event ma
 Let's first change our tests to showcase the behaviour we want the registry to exhibit. Open up `test/kv/registry_test.exs` and change the existing `setup` callback to the one below, then add the new test:
 
 ```elixir
-defmodule Forwarder do
-  use GenEvent
+  defmodule Forwarder do
+    use GenEvent
 
-  def handle_event(event, parent) do
-    send parent, event
-    {:ok, parent}
+    def handle_event(event, parent) do
+      send parent, event
+      {:ok, parent}
+    end
   end
-end
 
-setup do
-  {:ok, manager} = GenEvent.start_link
-  {:ok, registry} = KV.Registry.start_link(manager)
+  setup do
+    {:ok, manager} = GenEvent.start_link
+    {:ok, registry} = KV.Registry.start_link(manager)
 
-  GenEvent.add_mon_handler(manager, Forwarder, self())
-  {:ok, registry: registry}
-end
+    GenEvent.add_mon_handler(manager, Forwarder, self())
+    {:ok, registry: registry}
+  end
 
-test "sends events on create and crash", %{registry: registry} do
-  KV.Registry.create(registry, "shopping")
-  {:ok, bucket} = KV.Registry.lookup(registry, "shopping")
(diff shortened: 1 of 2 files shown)


getting-started/mix-otp/genevent.markdown
```diff
@@ -73,30 +73,34 @@ In order to emit events, we need to change the registry to work with an event ma
 Let's first change our tests to showcase the behaviour we want the registry to exhibit. Open up `test/kv/registry_test.exs` and change the existing `setup` callback to the one below, then add the new test:
 
 ```elixir
-defmodule Forwarder do
-  use GenEvent
+defmodule KV.RegistryTest do
```
josevalim: Well, this can be misleading because that's not the whole test file, right? We are supposed to have other test cases in there at this point?

eksperimental:
Also @josevalim, in the test.exs file

defmodule Formatter, should be inside `KV.RegistryTest

josevalim:
It is not supposed to matter. :)

getting-started/mix-otp/genevent.markdown
eksperimental: you are right, i completely overlooked at that line:
> and change the existing `setup` callback to the one below, then add the new test:

I think what i have been doing was replacing the tests, not updating them

eksperimental:
I reverted the changes and rebased.

josevalim:

eksperimental:
:wink:

PR #468, opened 2015-02-15
Add script to run jekyll server

eksperimental:
I find a bit longish
`$ bundle exec jekyll serve`

`$ ./serve` will do.

If merged, i will update [README.md]


josevalim:
I think it is just better to tell people to use the jekyll script. After all, there are other commands there that may be useful. :)

PR #469, opened 2015-02-15
jekyll-toc: fix css issue

eksperimental:


eksperimental:
sorry, jose, just give me a few minutes, i will update something that disables the animation completely when duration is 0. i will rebase and let you know.

josevalim:
Sure, no worries! Just let me know!

eksperimental:
fixed.

js/toc/toc.js
```diff
@@ -64,21 +64,40 @@
...
+          var top_element = ( $(settings.backToTopSelector).length ) ? $(settings.backToTopSelector).first() : $('html, body');
           var top_element_id = $(top_element).attr('id');
-          window.location.hash = ( typeof top_element_id === "undefined" ) ? '' : top_element_id;
+          update_hash ( (typeof top_element_id === "undefined") ? '' : top_element_id );
+          if ( settings.backToTopDuration > 0 || window.location.hash == '' ) {
+            $('html, body').animate({scrollTop: top_element.offset().top}, settings.backToTopDuration);
+          } else {
+            // force update
+            console.log("force update");
+            window.location.hash == window.location.hash
```
josevalim: Should this be `=`? Do we still need to log?

css/style.css
```diff
@@ -914,9 +914,9 @@ li.image {
 }
 
 /* extending jekyll-toc */
-.icon, .icon span {
+.icon {
   font: normal normal normal 16px/1 icons !important;
-  display: inline-block !important;
+  /* display: inline-block; */
```
josevalim: We can remove the comment here too!

js/toc/toc.js
```diff
@@ -64,21 +64,40 @@
...
 
+    function update_hash(hash) { 
+      if(history.pushState) {
+        history.pushState(null, null, '#'+hash);
+      } else {
+        location.hash = '#'+hash;
+      }
+    }
+
     function animate_link_here(header_id) {
```
josevalim: Can't we just remove the animation altogether from the plugin? This way we don't need update_hash because we will stop preventing the default.

josevalim:
I have merged this to fix the website but I have dropped some other comments. I have also noticed that the nested lists in the toc are messed up:

screen shot 2015-02-15 at 16 30 15

I am thinking we should simplify the toc plugin and get rid of the animation code altogether. Thoughts?

eksperimental:
so sorry about this, after i posted the message i realized i had unstaged commits, so i forced push, but you were too quick

js/toc/toc.js
eksperimental: well, the thing is that i'm planning on releasing this code, so other people can use it and we still can run the same code in the website, let me see if if just by settings i can force the natural behaviour, and if not i will remove it alltogether

js/toc/toc.js
eksperimental: yeah, that was not the version I ment to push. ;-(

josevalim: Ok, no worries, just send another PR. But I think it would be cool to completely get rid of the animation since we are not using it. Or at least change the code to not prevent default if there is no animation. :)

josevalim: Do you want me to revert those changes so it is easier for you to push the new ones?

josevalim:
I have reverted this and the original pull request so we can tackle this carefully. Thanks!

js/toc/toc.js
eksperimental: @josevalim yes, please, i tried to rebase and it wouldn't merge, so please do so. thanks

PR #470, opened 2015-02-15
Table of Contents: Animations removed

eksperimental:


eksperimental:
Let's see how it goes this time,
I have removed everything that had to do with animation., and it's pretty much plain anchors, in javsacript, with the execption of the title, that you can click on it (that 's attached the onclick event).

I have added a fallback to load jQuery locally, in case there's no internet, based on https://github.com/h5bp/html5-boilerplate

too bad I cannot update the code to test live, but the website only runs on the root directory.

josevalim:
Merged manually (skipped the jquery and demo bits). Thank you!

eksperimental:
Thank you Jose,

1. Internet was gone here, so I wasn't able to work on the site without loading jquery. So I think it won't hurt to have that option.
2. Demo I left it there as a way for other developers to see what's the contents of the fonts, in case they want to reuse them.

PR #477, opened 2015-02-27
Move TOC bellow introductory text

eksperimental:


josevalim:
I would like to keep the TOC where they are as there is no guarantee on the length of the first section. Thank you!

eksperimental:
sure. no worries

PR #480, opened 2015-02-27
bring back numbers to TOC

eksperimental:
as proposed in #477


eksperimental:
Also, please let me know if you are interested in having the changed I made in  42402a7
but I don' t know if they are necessary since TOC is at the top of the page.

PR #483, opened 2015-03-01
Adds support for Continous Integration

eksperimental:
I've been wanting to pull this one out for a while, and finally I did it.
This PR adds continuous integration for the Elixir Website,
It runs two different kinds of validation,

**html5validator**
[The W3C Validator](http://validator.w3.org/) developed by W3C and Mozilla.
It uses a precompiled package for python ([html5validator](https://github.com/svenkreiss/html5validator)).
This is the de-facto validator for html5, it will only check the markup validity

**html-proof**
[html-proof](https://github.com/gjtorikian/html-proofer) is a Ruby Gem. It will do all kind of validations, in addition to validate markup, it will check for favicons, external scripts being loaded, broken links and also internal broken links, images (internal and external). Anything that can break overtime (specially because we rely on external sources, ie, images, fonts, libraries), this will check for that, every time we make a commit.

The reality is that our website was not passing neither of both validations due to bugs [[1](https://github.com/svenkreiss/html5validator/issues/4),[2](https://github.com/gjtorikian/html-proofer/issues/190),[3](https://github.com/gjtorikian/html-proofer/issues/191)]  in these validators, and due nature of the website having hacky code to fix issues with old browsers. So that's why i created a script that will filter the errors in STDERR, and let us run the validators in Travis.

I wish we could [validate the css code](http://jigsaw.w3.org/css-validator/), but right now it's not possible, because the css it too hacky to support IE6 i think.

Links to the elixir documentation within the website have been disabled, because the /docs/ folder it not hosted within this project, We could configure html-proof to rewrite those relative links to absolute, but right now [a bug](https://github.com/gjtorikian/html-proofer/issues/192) prevented me from doing so. Anyway, I ran it locally, and I know it's all good.

You can see the travis build here: Usually it's been taking 1m 45sec, 
https://travis-ci.org/eksperimental/elixir-lang.github.com


.travis.yml
```diff
@@ -0,0 +1,14 @@
+language: ruby
+rvm:
+  - 2.1
+python:
+  - 2.7
+install:
+  - sudo pip install html5validator
```
josevalim: Do we need sudo? Because the builds with `sudo: false` are usually faster on travis-ci.

script/filter-lib.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
```
josevalim: Do we need the .sh extension on this file? Or does it mean it is not mean to be run directly?

josevalim:
This looks great, thank you!

script/filter-lib.sh
eksperimental: true, we don't need. i'll "push -f"

.travis.yml
eksperimental: I was getting errors without sudo.

eksperimental:
You are welcome José, I'm glad you liked it. :-D

script/cibuild
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
```
eksperimental: see need to add "set -e"
otherwise we will get false positives

eksperimental:
jose: could you please test script/cibuild on osx?, and tell me if it's all good.
also, if you can break a page, and see if it reports the error messages.
and if anybody else is using windows, that will be appreciated

.travis.yml
josevalim: Have you tried the version without sudo and then added `sudo: false` to the top of the .yml file?

eksperimental: same error:
```
error: could not create '/usr/local/lib/python2.7/dist-packages/html5validator': Permission denied
    Complete output from command /usr/bin/python -c "import setuptools, tokenize;__file__='/tmp/pip-build-JMP2ZA/html5validator/setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" install --record /tmp/pip-iH_ZLM-record/install-record.txt --single-version-externally-managed --compile:
```

josevalim: Ok, thanks for trying! <3

eksperimental: i found a workaround
https://github.com/travis-ci/travis-ci/issues/3139
```
sudo: false
install:
  - pip install --user html5validator
  - export PATH=$PATH:$HOME/.local/bin
```
it works, already tested

josevalim: Yay! I will try and merge this tomorrow!

eksperimental: :+1:
last commit pushed. all yours . and thanks for everythin

josevalim: Thank *you* for the fantastic work on ex_doc and on the website! <3

josevalim:
Should we close this one then?

eksperimental:
definitely

PR #493, opened 2015-03-09
Minor polishes in the "Crash course" guide

whatyouhide:
- Fix some capitalizations
- Uniform the style of not using spaces around curlies
- Add an example of an Elixir atom (`:""` to match the Erlang example of `''`)

Let me know if I can do anything else!


josevalim:

eksperimental:
I do remember long time ago, when I checked ran the examples on the erlang console, some were not working because they were missing the trailing "."

I have also a few imporvement I have written down somewhere for this page,  will update them as well

PR #495, opened 2015-03-24
Make explicit that client/server are 2 processes

edubkendo:
In a recent discussion in irc, someone new to Elixir was confused on this point. I wanted to make this what was going on a bit more explicit.


eksperimental:
this is a good point.
I will have to re-read the [processes](http://elixir-lang.org/getting-started/processes.html) section of the guide, but i think it will be good to include something like this there.

one of the things that was not clear to me, was that the only way to communicate from a spawned process back to the sender process is via message passing.
most of the times, they spawn a process, and what they do from this process is IO.puts the information,  but there it is being omitted that a value can be returned by sending a message to the sender process id.
It took me a while to figure this out

PR #504, opened 2015-04-08
Fix function call example

aerosol:


josevalim:

whatyouhide:
Are we sure it's not confusing if we show a call to `Orddict.new`, given that Elixir has no `Orddict` module?

josevalim:
Let's use `Map` (Elixir) and `maps` (Erlang) modules?

whatyouhide:
:+1:

Issue #511, opened 2015-04-26
Example for KV.Bucket needs a HashDict not a list


arthurcolle:
http://elixir-lang.org/getting-started/mix-otp/genserver.html

```
iex> Agent.start_link(fn -> [] end, name: :shopping)
{:ok, #PID<0.43.0>}
iex> KV.Bucket.put(:shopping, "milk", 1)
:ok
iex> KV.Bucket.get(:shopping, "milk")
1
```

I've been working through the examples and I couldn't get the above to work because I think the Agent needs to be started as such:

```
 Agent.start_link(fn -> HashDict.new end, name: :shopping)
```

The above code gave me this error, 

```
iex(2)> KV.Bucket.put(:shopping, "milk", 1)
** (EXIT from #PID<0.80.0>) an exception was raised:
    ** (FunctionClauseError) no function clause matching in HashDict.put/3
        (elixir) lib/hash_dict.ex:40: HashDict.put([], "milk", 1)
        (elixir) lib/agent/server.ex:20: Agent.Server.handle_call/3
        (stdlib) gen_server.erl:607: :gen_server.try_handle_call/4
        (stdlib) gen_server.erl:639: :gen_server.handle_msg/5
        (stdlib) proc_lib.erl:237: :proc_lib.init_p_do_apply/3

```

Which makes sense because in KV.Bucket, the start_link function starts returns HashDict.new, not an array.

I'm pretty sure of this but let me know if I'm doing something wrong. Thanks

josevalim:
Yes, good catch! Please send a pull request? Thank you!

arthurcolle:
Pull request sent.

PR #512, opened 2015-04-26
Fixed mix-otp/genserver typo

arthurcolle:
As per our discussion here: https://github.com/elixir-lang/elixir-lang.github.com/issues/511

There is an error where list is used in the Agent.start_link KV example rather than a HashDict, as per the bucket.ex example.

First pull request! Exciting!!


josevalim:

PR #513, opened 2015-04-29
Change controlling process in the gen_tcp tutorial

glesica:
The example as it is currently designed will cause all connections to be
dropped if the acceptor crashes because, by default, ports (as returned
by `:gen_tcp.accept/1`) are tied to the process that created them. See
the Erlang docs:

  * http://www.erlang.org/doc/reference_manual/ports.html
  * http://www.erlang.org/doc/man/gen_tcp.html

Thanks to @tsloughter for helping me confirm it.


josevalim:

PR #516, opened 2015-05-03
Update section on starting applications

### LINKED ISSUE #466: update change in "mix run --no-start" in  Supervisor and Application
Initially reported here: https://github.com/elixir-lang/elixir/issues/3071

We should udpated the guide
http://elixir-lang.org/getting-started/mix-otp/supervisor-and-application.html#starting-applications

Where it says to run `iex -S mix run --no-start:`
that it should output
```iex
iex> Application.start(:kv)
{:error, {:not_started, :logger}}
```
in reality, logger is loaded.

see:  https://github.com/elixir-lang/elixir/issues/3071
for josé's explanation.

amarandon:
I tried to rearrange the narration so that it exposes the same features
while being consistent with the current implementation.

Fix #466

### DIFF
--- a/getting-started/mix-otp/supervisor-and-application.markdown
+++ b/getting-started/mix-otp/supervisor-and-application.markdown
@@ -102,23 +102,36 @@ iex> Application.start(:kv)
 {:error, {:already_started, :kv}}
 ```
 
-Oops, it's already started.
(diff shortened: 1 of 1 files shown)


eksperimental:
Tha you for looking after this

PR #536, opened 2015-06-04
Minor updates and typos fix in mix and otp guide (part two)

dbbrian:
Here's part two. That guide is really good!


josevalim:
Glad you enjoyed and thanks for the reviews!  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #537, opened 2015-06-05
Caps and code block language hint

dbbrian:
Stuff I started checking only halfway through the first pass.


josevalim:

Issue #521, opened 2015-05-15
Add Erlang in Action to the list of learning resources


josevalim:
http://www.erlang-in-anger.com/

PR #538, opened 2015-06-06
Add Erlang in Anger to the learning resources #521

dbbrian:
Hi! I think the other covers should be stored here too. This kind of “hot linking” make the site dependant on *five* other sites.


josevalim:
Moving the covers would be a great idea, PR please?

Thank you!  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

dbbrian:
Done!

PR #544, opened 2015-06-18
Remove potentially destructive comprehension example in Getting Started guide

dougal:
One of the examples in the comprehension Getting Started guide contained a potentially destructive operation on the filesystem.

This commit replaces that operation (File.rm!/1) with a function that poses no risk to the files of a user following along (File.stat!/1).


josevalim:
Good call!  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #551, opened 2015-07-03
Clarify location of KV.Bucket.Supervisor

ryanb:
Just a little thing. It isn't obvious where this module should be defined. I first tried putting it in `lib/kv/bucket.ex` but the lookup doesn't check there.


josevalim:
Thanks @ryanb! :heart: That's indeed the best location although it should work on any directory in lib because it is all compiled anyway (we don't map modules <-> file).

ryanb:
You're right, I probably didn't save that file or something silly when trying it in `lib/kv/bucket.ex`. BTW Elixir is awesome!

josevalim:
Thank you!

PR #553, opened 2015-07-04
Clarify that underscore would match if it were reached in execution.

wsmoak:
In the first example for `case`, we learn that underscore "would match any value".
It's then confusing to read in the third example that underscore "won't match".
While this is technically true in the example as written, (since x>0, the first pattern will match,) it's not consistent with the earlier statement.


josevalim:
:heart: :blue_heart: :green_heart: :purple_heart: :yellow_heart:

PR #554, opened 2015-07-05
Note that variables cannot be used as map keys.

wsmoak:
Reinforce that 'Maps allow any *value* as a key' means that variables are not allowed as keys.

This came up on irc and I thought the docs could be clearer.  Once you think about what "value" means in context, it's obvious, but the first time through it may not be.


josevalim:
Good catch! We definitely need this now although I will be glad to revert this when Elixir 1.2 comes out because we fill finally support it! :pray:

PR #558, opened 2015-07-09
Fix typo in Processes

wsmoak:
The phrase 'dozens of thousands' caught my eye.  How about this?


eksperimental:
I think what the author meant was "tens of thousands",....

josevalim:
Yes, let's please go with "tens or even hundreds of thousands"!

wsmoak:
That makes more sense!  Fixed.

josevalim:

eksperimental:
:+1: that reads even better

PR #559, opened 2015-07-09
Add Sublime Plugin to Code Editor Support list

wsmoak:


eksperimental:
Textmate bundle alwo supports sublim, i think it should be worth mentioning it

josevalim:

wsmoak:
@eksperimental the Textmate bundle is listed, and it's mentioned in the Readme for this one.  I think it does the syntax highlighting, and you need both.

eksperimental:
@wsmoak yes, what i meant is that make a reference to sublimetext, where Textmate Bundle is listed,
what do you think @josevalim ?

josevalim:
Yeah, we could call it Sublime / TextMate Bundle if it fits there.

PR #560, opened 2015-07-11
Don't double-escape backslash

henrik:
Currently rendered as a double backslash on http://elixir-lang.org/getting-started/io-and-the-file-system.html.

Also confirmed that GitHub renders it double if doubled, and renders it fine if single.


josevalim:

henrik:
\💖 :)

PR #563, opened 2015-07-13
update url to use https. standardize erlang.org/doc links to use www

eksperimental:
* update URL where the used http link redirects to https, and as well when a package is provided and the https link is available.
* standardize erlang.org/doc links to use www


josevalim:
Thank you! Let's do this except the erlang.org changes!

josevalim:

PR #565, opened 2015-07-13
Operator precedence

gwww:
Added page on operator precedence. It is reachable from the Getting Started / Basic operators page.


ericmj:
This should just be on the operators chapter imo, no need for a special chapter.

gwww:
Is that the Basic Operators chapter. Happy to put it there. What's the
consensus?

______
Sent using an itsy bitsy keyboard. Expect typos and brevity.

josevalim:
Thank you @gwww! we will discuss the best place to put it there so you don't have to move things around multiple times. :D

josevalim:
Merged manually, thank you!

PR #568, opened 2015-07-17
Fix case-cond-and-if document

yuku-t:
The previous code block says:

> ** (RuntimeError) undefined function: if/1


josevalim:

PR #579, opened 2015-08-04
Add slack signup form to important links

jeregrine:
@josevalim I think this is what you are looking for.


eksperimental:
It should also be included that link in this list.
http://elixir-lang.org/getting-started/introduction.html#asking-questions

stevedomin:
:heart: :heart: :heart:  @jeregrine

PR #585, opened 2015-08-19
The :shutdown signal is sent automatically

wsmoak:
Working through the docs, I read "To shutdown the registry, we are simply sending a `:shutdown` signal to its process when our test finishes."

However the code doesn't have any explicit :shutdown signal being sent.  

@antipax explained on irc that this happens automatically.


josevalim:

PR #589, opened 2015-09-02
Fix wrong error message when updating map for non exist key

larrylv:
``` elixir
iex(7)> map = %{:a => 1, 2 => :b}
%{2 => :b, :a => 1}
iex(8)> %{map | :c => 3}
** (KeyError) key :c not found in: %{2 => :b, :a => 1}
```


josevalim:
It depends on the Erlang version, unfortunately. We can merge this after Elixir 1.2 is out.

josevalim:
Merged into the upcoming 1.2 branch, thank you!

PR #591, opened 2015-09-06
try-catch-and-rescue "Variables scope": offer alternative

henrik:
Could be more helpful.

Symbols instead of bools in the example because I think it reads clearer here.


josevalim:

PR #593, opened 2015-09-07
Changed `C` code to proper `C` code.

kislayentronic:
Changed :- 
1. Length logic
2. Added semicolon


josevalim:
I think that was meant to be Javascript. I don't know why it was tagged as C. :) The main point is to have a pseudo code from imperative languages and I think that is close enough to JS.

Thanks for the PR anyway!

PR #594, opened 2015-09-08
Multiple generators in comprehension

kislayentronic:
@josevalim : I tried my best to give a representation of multiple generators but I am not sure whether its up to the mark you expected. Please do a review and let me know if something else has to be changed/added/removed.
I can see you are not available on IRC but there was one more point where I am not clear which way to represent i.e. *traversing*. A classical example of that could be 

```
iex> for user <- [%{name: "Me", phones: ["123", "345"]}], do: (for phone <- user.phones, do: phone)
```
or the same code can be written as (without nested output)

```
iex> for user <- [%{name: "Me", phones: ["123", "345"]}], phone <- user.phones, do: phone
```


josevalim:
The traversal case is already shown with the files example. This is great, thank you!

PR #595, opened 2015-09-09
Improves pin operator explanation and examples

hannesfostie:
I'm just getting started with Elixir and am reading through the Getting Started guides. The pin operator example in the pattern matching chapter got me confused. I've changed the explanation slightly, changed a variable name, and added a "you can also rewrite this as" piece of code to really drive the point home.

As I was asking about this on Slack, I understood it as soon as I compared it to interpolation. Would interpolation be a good term in this case? The way I see it, you're pretty much interpolating the tuple. I'm not sure if that's technically correct and would actually improve the guide or not.

Open to suggestions. These are some of the best docs I've read to date. Keep it up! <3

### DIFF
--- a/getting-started/pattern-matching.markdown
+++ b/getting-started/pattern-matching.markdown
@@ -129,17 +129,26 @@ iex> x = 2
 2
 ```
 
-The pin operator `^` can be used when there is no interest in rebinding a variable but rather in matching against its value prior to the match:
+The pin operator `^` should be used when you want to pattern match against an existing variable's value rather than rebinding the variable:
 
 ```iex
 iex> x = 1
 1
 iex> ^x = 2
 ** (MatchError) no match of right hand side value: 2
-iex> {x, ^x} = {2, 1}
+iex> {y, ^x} = {2, 1}
 {2, 1}
-iex> x
+iex> y
 2
+iex> {y, ^x} = {2, 2}
+** (MatchError) no match of right hand side value: {2, 2}
+```
+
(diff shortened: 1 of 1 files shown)


getting-started/pattern-matching.markdown
```diff
@@ -129,17 +129,26 @@ iex> x = 2
 2
 ```
 
-The pin operator `^` can be used when there is no interest in rebinding a variable but rather in matching against its value prior to the match:
+The pin operator `^` can be used when you have no interest in rebinding a variable but rather use its value in the match operation:
```
bbhoss: Might I suggest
```
The pin operator `^` should be used when you want to pattern match against an existing variable, rather than binding its value to something new.
```

hannesfostie: Seems better, but here's another tiny modification (improvement?):

```
The pin operator `^` should be used when you want to pattern match against an existing variable's value rather than rebinding the variable.
```

Will update code later on as I suspect maybe we'll get more people adding suggestions?

josevalim: I like this last one!

eksperimental:
sweet!,
@follmann has written something similar like your improved example.
https://gist.github.com/follmann/2ad1c7b2a5ac93469986
maybe there are more ideas to get from there.

hannesfostie:
@eksperimental thanks for the link!

@josevalim I just updated the code with that last example. Open to further suggestions of course, if any!

josevalim:

PR #602, opened 2015-09-17
Fixing typo (Snippet is JS, indicated is C)

kislayentronic:


eksperimental:
thanks @kislayentronic 

I personally find the paragraphs suffers from a severe case of _parenthesistis_.

I would  go with something like:

> Due to immutability, loops in Elixir – and in any other functional programming language – are written differently from imperative languages. For example, in an imperative language like JavaScript, one would write:

kislayentronic:
@eksperimental : Fixed it as per your suggestion :) .

eksperimental:
:+1:  sweet!

PR #607, opened 2015-09-28
Add PhoenixScreencasts to the learning page

gjaldon:
Hi there!

[Phoenix Screencasts](http://www.phoenixscreencasts.com) has been around for a few months now and thought it would be great to add it here for those looking to use Elixir for web development.


josevalim:
Thanks @gjaldon but I don't want to list screencasts that are specific to a particular library or framework inside Elixir. With books starting to come out for Phoenix, they should start looking at learning resources too, that would be a good place to list it.

gjaldon:
Ok copy that. Will try my luck the Phoenix books.

PR #609, opened 2015-09-29
Fix Code of Conduct markdown link escape chars

kkirsche:
Escape `_`'s in the Code of Conduct link so that it doesn't use italics


josevalim:

PR #611, opened 2015-09-29
Clarify Getting Started > 8. Modules

endersstocker:
If the previous instructions were followed verbatim, some output would not be displayed. Additionally, `iex` would crash upon reaching `UndefinedFunctionError`.


josevalim:

PR #618, opened 2015-10-13
try-catch-and-rescue: skipping error struct

henrik:
Wasn't clear to me that you didn't have to provide the error struct.


henrik:
Thanks, José! :star2: :star:

PR #619, opened 2015-10-13
Add a short documentation for `use`

weppos:
I was reading the "Getting started" guide and one of the things I noticed is that the section where `require`, `alias` and `import` are documented doesn't include any explanation for `use`. In fact, `use` is not mentioned at all in the entire getting started guide.

I know [`import`, `require` and `alias` are `SpecialForm`](http://elixir-lang.org/docs/v1.1/elixir/Kernel.SpecialForms.html) whereas `use` is [defined as `Kernel` macro](http://elixir-lang.org/docs/v1.1/elixir/Kernel.html#use/2), however it's not uncommon to see new users confused by this specific keyword: see [here](http://stackoverflow.com/questions/28491306/elixir-use-vs-import) and [here](http://stackoverflow.com/questions/29097918/elixir-what-does-the-use-keyword-do).

This branch takes bits and pieces from the answers above and the [`use/2` documentation](http://elixir-lang.org/docs/v1.1/elixir/Kernel.html#use/2), and adds a new section in the "getting started" to explain the purpose of `use`.

What do you think?

### DIFF
--- a/getting-started/alias-require-and-import.markdown
+++ b/getting-started/alias-require-and-import.markdown
@@ -118,6 +118,40 @@ In the example above, the imported `List.duplicate/2` is only visible within tha
 
 Note that `import`ing a module automatically `require`s it.
 
+## `use`
+
+`use` allows you to use a module in the current context. It `require`s the given module and then calls the `__using__/1` callback on it allowing the module to inject some code into the current context.
+
+```elixir
+defmodule Example do
+  use Feature, option: :value
+end
+```
+
+is compiled into
+
+```elixir
+defmodule Example do
+  require Feature
+  Feature.__using__(option: :value)
+end
+```
+
+For example, in order to write tests using the ExUnit framework, a developer should use the `ExUnit.Case` module:
+
+```elixir
+defmodule AssertionTest do
+  use ExUnit.Case, async: true
+
+  test "always pass" do
+    assert true
+  end
+end
+```
+
+By calling use, a hook called `__using__` will be invoked in `ExUnit.Case` which will then do the proper setup.
+
+
 ## Aliases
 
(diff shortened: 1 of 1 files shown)


josevalim:
THank you! @weppos the issue is that `use` is not necessarily lexically scoped as everything else. So maybe we could add this to the end of the chapter and start with something like you described: "Although not a directive, `use` is a macro tightly related to require, after all [current description]". What do you think?

weppos:
Fine for me. Changing it now.

weppos:
Done. Give it a look at let me know if it works for you.

getting-started/alias-require-and-import.markdown
```diff
@@ -118,6 +118,7 @@ In the example above, the imported `List.duplicate/2` is only visible within tha
 
 Note that `import`ing a module automatically `require`s it.
 
+
```
josevalim: Extra line.

getting-started/alias-require-and-import.markdown
```diff
@@ -192,3 +193,38 @@ end
 As we will see in later chapters, aliases also play a crucial role in macros, to guarantee they are hygienic.
 
 With this we are almost finishing our tour about Elixir modules. The last topic to cover is module attributes.
```
josevalim: This should be the last line of the chapter. :)

josevalim:
I have added two more comments and we are good to go. :)

getting-started/alias-require-and-import.markdown
weppos: It was probably added in the cut&paste. Removing it...

weppos:
Here we go. Thanks for your feedback.

josevalim:

PR #622, opened 2015-10-16
Clarified some stuff in the intro section

vikeri:
Hopefully made it somewhat more beginner friendly.

* Added disclaimer that terminal commands are for unix
* Updated the terminal commands to create, edit and run an elixir file

### DIFF
--- a/getting-started/introduction.markdown
+++ b/getting-started/introduction.markdown
@@ -48,17 +48,20 @@ It seems we are ready to go! We will use the interactive shell quite a lot in th
 
 ## Running scripts
 
-After getting familiar with the basics of the language you may want to try writing simple programs. This can be accomplished by putting Elixir code into a file and executing it with `elixir`:
+After getting familiar with the basics of the language you may want to try writing simple programs. This can be accomplished by putting the following Elixir code into a file:
+
+```elixir
+IO.puts "Hello world from Elixir"
+
+```
+
+ Save it as `simple.exs` and execute it with `elixir`:
 
 ```bash
-$ cat simple.exs
-IO.puts "Hello world
-from Elixir"
-
 $ elixir simple.exs
-Hello world
-from Elixir
+Hello world from Elixir
 ```
+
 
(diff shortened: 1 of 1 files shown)


eksperimental:
Whats the point of using touch? If we are using > later

josevalim:
Thank you! What if we change to not use terminal commands then? Just say: "imagine you have a file named this with these contents.... you can run it with ..." and that's it? Great catch.

vikeri:
I added `touch` to clarify the process: 
1. Create file
2. Edit the file
3. Run file.

But it could of course be removed.

I still think that it is good to make people use the terminal both because this part of the tutorial should be agnostic to tooling like editor choice etc. and also because if you don't already know these terminal commands it is good to get exposed to them.

josevalim:
> if you don't already know these terminal commands it is good to get exposed to them.

There are many other important skills that I'd say a developer needs to get exposed to. We could for example make people acquainted big-O notation, that's not the point of the this tutorial though. :) Specially when the commands are OS specific.

If you could amend your PR to not rely on the terminal commands, we would really appreciate it. Otherwise one of us can get to it later. :) Thank you!

vikeri:
Sure I'll update it :) but just to clarify (no WIndows computer nearby); how do I run elixir on windows? `elixir.exe simple.exs` or just `elixir simple.exs` or some other variant? And should I add hints that if elixir can't be found that it must be added to the path: http://elixir-lang.org/install.html#setting-path-environment-variable

josevalim:
`elixir simple.exs` should be enough. Don't worry about elixir not being on the path given it is covered in the install guide. :)

vikeri:
@josevalim Updated PR now

josevalim:

getting-started/introduction.markdown
```diff
@@ -48,18 +48,21 @@ It seems we are ready to go! We will use the interactive shell quite a lot in th
...
-```bash
-$ cat simple.exs
-IO.puts "Hello world
-from Elixir"
+```elixir
+IO.puts "Hello world from Elixir"
+
+```
 
+ Save it as `simple.exs` and execute it with `elixir`:
```
eksperimental: Wouldnt it be concenient as well to link here to a simple.exs file so people can download directly

PR #625, opened 2015-10-22
Fix extra word typo

danila:
Remove an extra "instead" word in the GenServer chapter of Getting Started


josevalim:

PR #627, opened 2015-10-26
Added more explanation to agent text

vikeri:
* Explained the arguments of the functions update and get
* Added note that Agents only store state and not logic


josevalim:
Thank you! Can you please keep the markdown text in organized paragraphs, similar to before?  Just  to keep it aligned with the remaining structure of the text, thank you!

vikeri:
Sure! Like that?

josevalim:

PR #629, opened 2015-11-07
Fix `File.rm_rf` arity.

ags:
`File.rm_rf` has an arity of 1 (https://github.com/elixir-lang/elixir/blob/v1.1.1/lib/elixir/lib/file.ex#L836),
but the getting started guide lists it as 2.


josevalim:

PR #631, opened 2015-11-09
Add Integer.is_even/1, Integer.is_odd/1 to functions allowed in guards

eksperimental:


josevalim:
This comes every once in a while. The goal is to cover only the ones in
Kernel.:)

-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

eksperimental:
true, we already talked about it here: https://github.com/elixir-lang/elixir-lang.github.com/issues/499#issue-66270596
I completely forgot about it

eksperimental:
Would it be worth mentioning somewhere that this is not a full list that comes with Elixir.
because actually, that's  the impression I had, even though it is not mentioned anywhere, only
> Additionally, users may define their own guards. For example, the Bitwise module defines guards as functions and operators...

PR #634, opened 2015-11-13
Mandatory function head

webengineer-max:
Body-less function head is required (not optional) for multiple clauses
with defaults. Without the head following error raised by compiler
```
== Compilation error on file lib/mnesia/key_gen.ex ==
** (CompileError) lib/mnesia/key_gen.ex:18: def slug/3 has default values and multiple clauses, define a function head with the defaults
    lib/mnesia/key_gen.ex:18: (module)
    (stdlib) erl_eval.erl:669: :erl_eval.do_apply/6
```

### DIFF
--- a/getting-started/modules.markdown
+++ b/getting-started/modules.markdown
@@ -219,7 +219,7 @@ hello
 :ok
 ```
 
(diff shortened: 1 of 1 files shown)


josevalim:

PR #635, opened 2015-11-16
Update note about required Elixir version to match the Introduction guide

carlosantoniodasilva:
The Introduction guide sets Elixir version 1.0.0 as the minimum requirement, so the Mix Introduction guide should match the same required version - it's likely what users will have installed anyway.


josevalim:

PR #636, opened 2015-11-22
Use MIX_ENV=test when starting `bar` for the tests

kivikakk:
Otherwise the tests can fail because the compiled BEAM files are different (and the anonymous functions won't run across nodes).


josevalim:

PR #637, opened 2015-11-24
Update fedora install instructions

ZJvandeWeg:
As `yum` is now replaced with `dnf`, the install instructions where not up to date.

It might be considered to remove the fedora 17 section all together, as only Fedora 21 packs it and older versions are [not maintained anymore.](https://fedoraproject.org/wiki/End_of_life?rd=LifeCycle/EOL). Fedora 21 EOL is in less than a week, december 1st.


josevalim:

PR #642, opened 2015-12-01
docs(install.md) change location for erlang package

Gazler:
The previous link was 404ing.


Gazler:
Old: https://www.erlang-solutions.com/downloads/download-erlang-otp
New: https://www.erlang-solutions.com/resources/download.html

PR #645, opened 2015-12-07
Modify anon func example to better match others

acook:
The existing version is syntactically valid, but breaks with the usage of parens in all the other examples.


josevalim:
I would keep it as is though because we rarely use parenthesis in function heads (the same way we rarely use parens on case heads). Thank you!

PR #647, opened 2015-12-21
Fix typo on `Mix.Config.config`

yzyzsun:


whatyouhide:
It's not a typo. That code is using the [`config/3` macro](http://elixir-lang.org/docs/stable/mix/Mix.Config.html#config/3) instead of `config/2` (even if in this instance, the existing one and the one you wrote do basically the same thing).

yzyzsun:
@whatyouhide Oops, you are right.

PR #649, opened 2015-12-25
Fix example exceptions for structs

dgarlitt:
Updated the following example exceptions for structs in the
getting-started guide:
 - when using pattern matching to update a field that doesn't exist in a
   struct
 - when trying to directly access a field in a struct
 - when trying to use the Dict.get method to access a field in a struct


josevalim:
Merged in master, thank you!

PR #651, opened 2016-01-03
[650] renamed markdown to fix broken link

sandeshsoni:
page i am talking about is -> Getting-started -> Mix and OTP -> Docs, tests and with

I have manually checked,
and confirm the broken link is fixed.

I have a question.

Why does the title of this page say *Docs, tests and with*
while it talks about *Doctests, with, with and Running commands* ?

I am new to Elixir.


josevalim:
Thank you! We cannot include everything we talk about in the title so we picked the most important ones. :)

PR #655, opened 2016-01-07
supervisor-and-application race condition - there is no way to handle…

tomjoro:
… in this chapter so a sleep (hack) is used and with a note to user. I can't think of any better solution, other than a spin-wait for the bucket to disappear. Otherwise you'd need to monitor the supervisor, or have it send another message when the DOWN message has been handled.

### DIFF
--- a/getting-started/mix-otp/supervisor-and-application.markdown
+++ b/getting-started/mix-otp/supervisor-and-application.markdown
@@ -236,8 +236,10 @@ In other words, we want the registry to keep on running even if a bucket crashes
     # Stop the bucket with non-normal reason
     Process.exit(bucket, :shutdown)
 
-    # Do a sync to ensure the registry processed the down message
-    _ = KV.Registry.create(registry, "bogus")
+    # The :DOWN message is handled asynchronously, so we don't  know if it has been handled yet, so give it a chance to run...
+    # We will fix this race condition in the next chapter
+    :timer.sleep(100)
+
     assert KV.Registry.lookup(registry, "shopping") == :error
   end
 ```


josevalim:
Thank you. I have also pushed a fix that relies on monitor + DOWN. :)

tomjoro:
Your fix is to the ETS chapter - and that works. But my PR is for the supervisor chapter. If you just do the work up the end of the Supervisor chapter then the test cases will fail (which is what happened to me and others) so some correction is needed in this chapter also.

tomjoro:
sorry, my mistake.. now I see your change. But I'm still not sure if waiting for a message in another monitor will actually guarantee the message is processed in the Registry, but it does work.

josevalim:
@tomjoro it guarantees. All down messages are guaranteed to be delivered at the same time. :)

tomjoro:
oh, that's handy :) ..  does that mean then that the code to create bogus bucket should be put back into the example, because the code in Registry that handles must run to completion before we try to check if the bucket exists (because it is in a Map managed by Registry).

tomjoro:
The registry lookup is also a message, so it would be queued after the :DOWN, and hence synchronized (I assume handle_info uses the same queue) Thanks! I get it now. It'll work always...

josevalim:
> The registry lookup is also a message, so it would be queued after the :DOWN, and hence synchronized (I assume handle_info uses the same queue) Thanks! I get it now. It'll work always...

You got it! :D

tomjoro:
IMO that's a super important feature (especially for testing) and maybe deserves mentioning.. "All :DOWN messages are guaranteed to be delivered at the same time". It seems a common testing scenario when doing third party (can I call that?) monitoring like this. Some of my Erlang friends didn't know this.

josevalim:
@fishcakez do you agree we should make it clear in the guides are DOWN messages are delivered at once?

fishcakez:
That guarantee does not exist, DOWN are sent async, one after other and it costs reductions as normal AFAIK.

josevalim:
Damn it. Are links the ones that are guaranteed? Is there any way to guarantee the process may have a received a DOWN notification? Or are they delivered in order (the first to monitor is the first to receive it)?

fishcakez:
Links are the same as monitors. The only guarantee is links are sent before monitors: https://docs.google.com/presentation/d/1dRUQ1jC0NXRyM-6WJWo15n-3hSZ1p7F8fl3X4bdhis8/pub?start=false&loop=false&delayms=60000#slide=id.g34cd3d110_0282

josevalim:
Thanks, I guess we cannot rely on any ordering guarantee either. I will think about it more.

PR #660, opened 2016-01-09
Fixing typo in dependency paragraph

devonestes:
The code example of an external dependency was for `{:plug}, "~> 1.0"}`, but the text said that the dependency referred to the latest version of Plug in the 0.5.x version series, which I think is old and should be updated to reference the 1.0 version series.


josevalim:

PR #661, opened 2016-01-11
Set width of the kv-observer image to fit better in the desktop

hugobarauna:
The way it its, the image's width is larger than its `<p>` container:

image

I just added a `width=640px` to the image so that it fits into its container:

image


josevalim:
Can we instead add a `width="640"` attribute to the image? Otherwise the quality will be poor for those on retina displays.

hugobarauna:
@josevalim done

josevalim:

PR #664, opened 2016-01-16
fixed typo

riseshia:
fixed 'function stars a new process' to 'function starts a new process' :)


josevalim:
That's a stellar change!

PR #665, opened 2016-01-18
Unnecessary 'it' removed

meinac:
I've interested in elixir about couple of weeks ago and reading through some books and elixir-lang website. I've encountered with this while reading about ETS. I thought the first `it` is unnecessary and removed it.
By the way, thank you for this nice language.


josevalim:

PR #672, opened 2016-01-28
Update string guide

endersstocker:
Updated to indicate that `String` now passes all the tests in "The string type is broken" (as of Elixir 1.2).


josevalim:

PR #676, opened 2016-02-05
Set a default permalink for blog posts

cazrin:
Due to the recent jekyll changes I noticed that a permalink for each blog post had to be added. Rather than have that added to every blog post file, this `_config.yml` change provides a default permalink for all `post` types.


josevalim:

PR #681, opened 2016-02-12
Added a section about procedure-level `catch` blocks.

tangentstorm:
Someone asked what the "`catch` without `try`" form meant in the IRC channel, and I remembered being confused about it myself when I first saw it.


whatyouhide:
This actually works with `rescue` and `after`, so it may be worth to mention them as well :). For example:

```elixir
def my_fun() do
  do_stuff()
after
  release_resources()
end
```

josevalim:

josevalim:
I have amended locally, thank you!

PR #683, opened 2016-02-15
use real modules in crash course

mitchellhenke:
It's been misleading for a couple of people I've seen working through the crash course to reference modules that don't exist.

I wasn't sure what the best example would be, but I used `List` and `lists` since an example using `List` follows.


josevalim:

PR #686, opened 2016-02-22
link to most recent (v1.2.3) Docs.zip

alxndr:
Noticed that this was linking to a slightly out-of-date version. Is there a better way of keeping links like this up to date? (The `/docs/v1.2/.../` links below do redirect to the most recent patch version.)


josevalim:
Good call, I will figure out a better mechanism.

josevalim:

PR #694, opened 2016-03-03
Fix a Problematic Link in the Mix & OTP Guide

JEG2:
The presence of a `|` character in a Markdown link on the _Task and gen_tcp_ page was causing the paragraph it's in to render as a table:

task_and_gen_tcp_-_elixir

Also, the link didn't properly function for me.

I've escaped the entities in it which fixes the rendering:

task_and_gen_tcp_-_elixir

This also makes the link work in Chrome and Safari.  I was unable to find a way to make Firefox accept it.


josevalim:
Amazing, haha!

josevalim:

Issue #696, opened 2016-03-11
A link to the repository on every page (in the footer?)


lexmag:
I think we should have something like this on every page:

> This website is [open source](https://github.com/elixir-lang/elixir-lang.github.com). See a typo? Please, [open an issue](https://github.com/elixir-lang/elixir-lang.github.com/issues).

Right now we mention this only once:  at [the beginning of getting started](http://elixir-lang.org/getting-started/introduction.html).

Inspired by [InfluxDB docs](https://docs.influxdata.com/influxdb/v0.10/query_language/functions/).

josevalim:
Instead of "open an issue", I would link to the source code on Github so folks can press edit and submit the change. :)

lexmag:
Closed in #697 and improved in #698.

PR #697, opened 2016-03-11
Add a "Edit on GitHub" link to the footer

whatyouhide:
I love what @lexmag and @josevalim discussed in #696. How does this implementation look? I haven't tested the proper link locally because GH injects the `site.github` data when it builds the site.

### DIFF
--- a/_includes/bottom.html
+++ b/_includes/bottom.html
@@ -2,6 +2,10 @@
 
       <div class="clear"></div>
 
+      <a id="edit-on-github" href="{{ site.github.repository_url }}/edit/{{ site.branch }}/{{ page.path }}">
+        Something is wrong? Edit this page on GitHub.
+      </a>
+
       <div id="copyright">
         &copy; 2012-2016 <a href="http://plataformatec.com.br/">Plataformatec</a>. All rights reserved.
       </div>
--- a/css/style.css
+++ b/css/style.css
@@ -765,6 +765,12 @@ li.image {
   padding: 0 0 10px;
 }
 
+#edit-on-github {
+  display: block;
+  text-align: center;
+  font-size: 12px;
+}
+
 /*  Media Queries (mobile browsing)
 ----------------------------------------------------- */
 /* Tablet (portrait) */


_includes/bottom.html
```diff
@@ -2,6 +2,10 @@
 
       <div class="clear"></div>
 
+      <a id="edit-on-github" href="{{ site.github.repository_url }}/edit/{{ site.branch }}/{{ page.path }}">
+        Edit this page on GitHub
```
lexmag: No sure but maybe "If you see a typo, please, <a id=..." even more encourage people to edit. ¯\\_(ツ)_/¯

josevalim: What about: "Something is wrong? Edit this page on Github."

lexmag: Even better. :+1:

whatyouhide: Agreed, @josevalim's message sounds great. Amended the commit!

_includes/bottom.html
```diff
@@ -2,6 +2,10 @@
 
       <div class="clear"></div>
 
+      <a id="edit-on-github" href="{{ site.github.repository_url }}/edit/{{ site.branch }}/{{ page.path }}">
+        Something is wrong? Edit this page on GitHub.
```
josevalim: I just noticed we are including this on all pages. I wonder if we should instead include this only on the Getting Started guide?

josevalim: The rationale is that we could move it up in the guides (before or after the pagination control) instead of the absolute bottom. Thoughts?

whatyouhide: I thought about this, but every page can be edited and every page can contain typos/other kind of mistakes, so I went with this. I can move this in the Getting Started guide of course :)

josevalim: Btw, the link does not work, probably because some pages are .html and others are .md. Another reason for moving to the getting started guide only I believe. :)

whatyouhide: Ah, that's a shame. Then yes, let's move to the Getting Started only, PR coming right away.

whatyouhide:
@josevalim do you think something like this would work style-wise?

screen shot 2016-03-11 at 5 26 55 pm

_includes/bottom.html
lexmag: Also it looks better when only "Edit this page on GitHub." part is a link but "Something is wrong?" is a text. :bowtie:

josevalim:
Agreed with @lexmag . Let's make only the action a lnik. Below the footer looks nicer.

PR #704, opened 2016-03-17
Add return value for list example in getting started for basic types

amorphid:
changes this...

```
iex> list = [1,2,3]
iex> hd(list)
1
```
...to this...
```
iex> list = [1,2,3]
[1, 2, 3]
iex> hd(list)
1
```

### DIFF
--- a/getting-started/basic-types.markdown
+++ b/getting-started/basic-types.markdown
@@ -257,6 +257,7 @@ Throughout the tutorial, we will talk a lot about the head and tail of a list. T
 
 ```iex
 iex> list = [1,2,3]
+[1, 2, 3]
 iex> hd(list)
 1
 iex> tl(list)


whatyouhide:
There are many many places in the documentation where this pattern is used, because it can quickly become annoying and clutter-y if you always write the return value of each expression in a `iex>` prompt. In doctests, it's as common to avoid return values because often you don't know what the return value will be (think of a dynamic pid). For these reasons, I tend to be :-1: towards this change.

amorphid:
I don't have a strong opinion on this.  Just took two seconds to submit a fix that looked like an oversight :)

josevalim:
Thank you @amorphid! Let's try skip this change for now but I completely agree it may look as an oversight. If others feel this is indeed confusing, we can merge future changes. :)

PR #707, opened 2016-03-19
Make correction in “Sigils”

endersstocker:
The macro `sigil_r` was incorrectly identified as a function. One could have merely changed the word _function_ to _macro_. However, the revelation that it is a macro and not a function could pose a distraction and lead to confusion ahead of the final paragraph of the page. Consequently, I chose to remove the word _function_.


whatyouhide:
:clap:

josevalim:

PR #714, opened 2016-03-29
Add reference to capture syntax for anonymous functions

chuckus:
Add reference to capture syntax for anonymous functions so that when going through 'Getting started' in order, you don't get curious at Keywords and maps - Nested data structures, which uses the capture syntax.

### DIFF
--- a/getting-started/basic-types.markdown
+++ b/getting-started/basic-types.markdown
@@ -233,6 +233,8 @@ iex> x
 42
 ```
 
+Capture syntax `&()`can also be used for creating anonymous functions. This type of syntax will be discussed in Chapter 8.
+
 ## (Linked) Lists
 
 Elixir uses square brackets to specify a list of values. Values can be of any type:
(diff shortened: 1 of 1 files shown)


getting-started/basic-types.markdown
```diff
@@ -233,6 +233,8 @@ iex> x
 42
 ```
 
+Capture syntax `&()`can also be used for creating anonymous functions. This type of syntax will be discussed in Chapter 8.
```
whatyouhide: > **The** capture syntax [...]

? Also, we could make "Chapter 8" a link to chapter 8 and `&()` a link to http://elixir-lang.org/docs/stable/elixir/Kernel.SpecialForms.html#&/1. Wdyt?

chuckus:
@whatyouhide Agree with changes, followed the syntax already on the page for the link to the Elixir kernel docs.

whatyouhide:
Looks good @chuckus, great work!

josevalim:

PR #715, opened 2016-03-29
Fix windows set environment cmd

Steve-O-Cassels:
Great project, thank you!

The example for setting a session based environment variable for establishing your Elixir environment is incorrect. 

It uses the `-A` flag which [is for expressions](http://ss64.com/nt/set.html) like setting a variable to the result of '1 + 2'

If you try to use batch with the existing cmd using the /a flag, the output in windows cli is `0` - not set.  My revision will resolve this and is tested on Windows 8.1


josevalim:

Issue #727, opened 2016-04-18
Examples in "Getting Started: case, cond and if" don't match actual iex output.


rayhamel:
Elixir 1.2.4/Erlang 18.3/Mac OS X 10.11 El Capitan

On the ["case, cond and if" page](https://github.com/elixir-lang/elixir-lang.github.com/blob/master/getting-started/case-cond-and-if.markdown) of the "Getting Started" section, the following two code examples are shown as throwing `(RuntimeError) undefined function: if/1` in iex:
```
iex> is_number if true do
...>  1 + 2
...> end
** (RuntimeError) undefined function: if/1
```
and
```
iex> is_number(if true) do
...>  1 + 2
...> end
** (RuntimeError) undefined function: if/1
```
However, they actually throw `(CompileError) undefined function is_number/2` on my computer, as seen in the attached screenshot.

![screen shot 2016-04-17 at 7 00 29 pm](https://cloud.githubusercontent.com/assets/9465511/14590714/67a9351c-04cf-11e6-80ca-b37e72c5bb4a.png)

PR #729, opened 2016-04-18
Fix the output of some examples in the "Case, cond and if chapter"

whatyouhide:
This is a possibly temporary fix for #727. I think we may want to rework the examples some way so that they become clearer, but this is a solution to the problem we're having *now* :)


josevalim:

PR #730, opened 2016-04-18
Fix typo

formido:
Could be also fixed by: "Now all data types (including structs) in which we have not implemented the Blank protocol will be considered non-blank."


josevalim:

PR #731, opened 2016-04-18
Update syntax style on examples

alfakini:
Just some nits :)

### DIFF
--- a/getting-started/basic-types.markdown
+++ b/getting-started/basic-types.markdown
@@ -355,7 +355,7 @@ iex> list = [1|[2|[3|[]]]]
 This means accessing the length of a list is a linear operation: we need to traverse the whole list in order to figure out its size. Updating a list is fast as long as we are prepending elements:
 
 ```iex
-iex> [0 | list]
+iex> [0|list]
 [0, 1, 2, 3]
 ```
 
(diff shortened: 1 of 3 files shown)


getting-started/basic-types.markdown
```diff
@@ -355,7 +355,7 @@ iex> list = [1|[2|[3|[]]]]
 This means accessing the length of a list is a linear operation: we need to traverse the whole list in order to figure out its size. Updating a list is fast as long as we are prepending elements:
 
 ```iex
-iex> [0 | list]
+iex> [0|list]
```
whatyouhide: Maybe it's better if we keep this with spaces, given there's [a discussion in an Elixir PR](https://github.com/elixir-lang/elixir/pull/4507) around deciding on the `[x | y]` style once and for all, throughout all Elixir's codebase. :) Same applies to the other occurrences below.

josevalim:
@alfakini can you please send a PR that only takes the space from the right of `{` and left of `}`? The rest is correct as today.

getting-started/basic-types.markdown
eksperimental: I have a commit ready to send fixing all the spaces around the `|`,
I will send it soon

PR #732, opened 2016-04-19
Clearer, working examples for "case, cond and if" chapter

rayhamel:
Fix for #727 that once again clearly demonstrates the ambiguity in the arguments to `if/2`. Still learning Elixir, but I'm guessing the original examples with `is_number/1` didn't produce the same output as shown on the page because of how the compiler handles macros vs. regular functions.

Ref: Issue #727, PR #729 
screen shot 2016-04-18 at 11 34 10 pm

### DIFF
--- a/getting-started/case-cond-and-if.markdown
+++ b/getting-started/case-cond-and-if.markdown
@@ -251,7 +251,7 @@ One thing to keep in mind when using `do/end` blocks is they are always bound to
 iex> Integer.to_string if true do
 ...>  1 + 2
 ...> end
-** (CompileError) undefined function if/1
+** (CompileError) undefined function: if/1
 ```
 
 Would be parsed as:
@@ -260,7 +260,7 @@ Would be parsed as:
 iex> Integer.to_string(if true) do
 ...>  1 + 2
 ...> end
-** (CompileError) undefined function if/1
+** (CompileError) undefined function: if/1
 ```
 
 Which leads to an undefined function error as Elixir attempts to invoke `if/1`. Adding explicit parentheses is enough to resolve the ambiguity:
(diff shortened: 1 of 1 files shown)


getting-started/case-cond-and-if.markdown
```diff
@@ -248,28 +248,28 @@ iex> if true, do: (
...
-iex> is_number(if true) do
+iex> Integer.to_string(if true) do
 ...>  1 + 2
 ...> end
-** (CompileError) undefined function: is_number/2
+** (CompileError) undefined function: if/1
 ```
 
-which leads to an undefined function error as Elixir attempts to invoke `is_number/1`, but passing it *two* arguments (the `if true` expression - which would throw an undefined function error itself as `if` needs a second argument, the `do/end` block - and the `do/end` block). Adding explicit parentheses is enough to resolve the ambiguity:
+Which leads to an undefined function error as Elixir attempts to invoke `if/1`. Adding explicit parentheses is enough to resolve the ambiguity:
```
whatyouhide: I changed "Which" to "which" on purpose as we don't have a period on the last line of text above this

> Would be parsed as:

Actually, we should probably change that to "which" as well now that I think about it :)

josevalim:
Thank you @rayhamel!

@whatyouhide has already fixed this and I am getting exactly the same behaviour he described in the post. Do you get a different one when using `is_number`?

whatyouhide:
@josevalim he's probably getting the same behaviour as in the example I added yesterday, but by using `is_number/2` the `CompileError` mentions the focus is shifted away from `if`. With @rayhamel's fix (which I believe is just using a function - `Integer.to_string` - instead of a macro) the error is actually about the `if` (which is evaluated first being argument to a function). I believe this is better :)

josevalim:
@whatyouhide I think both are valid. It depends which one will be more common (and I believe yours is but I have no data to back it up). The important though is not exactly the scenario but understanding the precedence. That said, I will keep things as is.

Thank you anyway @rayhamel!  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

whatyouhide:
Yeah, they should convey the meaning of this anyways. :)

PR #742, opened 2016-05-05
the send command returns the tuple, not the #PID

tschmits:
http://elixir-lang.org/docs/stable/elixir/Kernel.html#send/2:

> Sends a message to the given dest and returns the message.

Which, in this case, is the tuple.


josevalim:

PR #744, opened 2016-05-10
Clarify semantics of `after`.

myronmarston:
See https://groups.google.com/forum/#!topic/elixir-lang-core/NWvFwUWo2QQ for what I was confused about.

### DIFF
--- a/getting-started/try-catch-and-rescue.markdown
+++ b/getting-started/try-catch-and-rescue.markdown
@@ -159,7 +159,7 @@ It is exactly this supervision system that makes constructs like `try/catch` and
 
 ## After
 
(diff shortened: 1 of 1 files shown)


getting-started/try-catch-and-rescue.markdown
```diff
@@ -172,6 +172,11 @@ iex> try do
 ** (RuntimeError) oops, something went wrong
 ```
 
+The `after` clause will be executed regardless of whether or not the
+tried block succeeds. Note, however, that if a linked process exits,
+this process will exit and the `after` clause will not get run. Thus,
+`after` provides only a soft guarantee.
```
josevalim: Given my other comment, what if we add something like: "However, given the opened file is also linked to the current process, it will always be closed if the current process crashes, independent of the after clause."

myronmarston: I pushed a change based on your suggestion.  Let me know what you think!

PR #745, opened 2016-05-12
Use %{} instead of Map.new in genserver docs

radar:
I am not sure why we're using `Map.new` here instead of `%{}`. On the previous page we're using `%{}`, but here we're using `Map.new` for apparently no reason at all.


josevalim:

PR #747, opened 2016-05-14
Fixing compile error

muaazsaleem:
Left over "def run(command)" in the code, makes compilation fail


josevalim:
@muaazsaleem that's a valid Elixir construct and it should not make compilation fail. It is useful when you want to define the function name and its arguments before implementing all of its clauses. :)

muaazsaleem:
Oh I must not have all the code right. Sorry to wast your time :(

josevalim:
No problem :)

PR #749, opened 2016-05-16
Update install.markdown

arcz:
I feel like current description suggests that it always happens that binaries are missing from path.


josevalim:

PR #753, opened 2016-05-27
Fix undefined function error in Agent.update & Agent.get examples

keiththomps:
When going through this section I ran into the following error:

undefined function

To make this work I went with the same structure used in the [Agent documentation](http://elixir-lang.org/docs/stable/elixir/Agent.html).


whatyouhide:
The problem in your code is that you're not declaring `map` as an argument to the `fn`. Can you try and do that?

keiththomps:
Yeah I just noticed that actually. I think github actually makes it a little *too* easy to open a pull request. Sorry about that.

PR #754, opened 2016-05-29
Update 'case, cond, and if' page

collinksmith:
It looks like something got mixed up here, as there's a confusing repetition. I think this is the clearest option to fix it. Alternatively, it could say something like

> `if` needs two arguments - the conditional expression and the `do/end` block.


josevalim:

PR #758, opened 2016-06-02
Change sentence in alias-require-and-import.markdown to use idiomatic English

emwalker:
This commit changes "With this we are almost finishing our tour about Elixir modules. " to "With this we have almost finished our tour of Elixir modules."


josevalim:

PR #766, opened 2016-06-10
Replace JavaScript example with C

aliismayilov:
Because in JavaScript one could write that example in imperative way, but would write that example as
```javascript
array.map(function(el) {
  return el * 2;
});
```

Let's just use pure imperative programming language as an example.


josevalim:

PR #773, opened 2016-06-13
Redirect old getting started URL

conroywhitney:
The Pragmatic "Programming Elixir" book references `http://elixir-lang.org/getting_started/1.html` which has been moved to `http://elixir-lang.org/getting-started/introduction.html`. This PR adds a `redirect_from` on the `introduction.markdown` page to ensure that readers of that book are able to use the URL in the book without getting a 404.

### DIFF
--- a/getting-started/introduction.markdown
+++ b/getting-started/introduction.markdown
@@ -1,6 +1,7 @@
 ---
 layout: getting-started
 title: Introduction
+redirect_from: /getting_started/1.html
 ---
 
 # {{ page.title }}


josevalim:

eksperimental:
I think https://github.com/conroywhitney/elixir-lang.github.com/commit/c6183c6daaf29b53b352e79792ca9207428d9976 should ve reverted. I think it was agreed to keep it like that as there are countless of webpages linking to the old urls.

josevalim:
@eksperimental I have decided to merge it back because this is the third or fourth time someone complains about this URL explicitly. Since it was the entrypoint into the guide, we shuold probably keep it in the long term.

eksperimental:
Im totally fine with it. What concerns me is that all the other old links were removed. What was the point in doing so? There are many sources still linking to the old urls

eksperimental:
Just to be clear. The commit that i mention should be reverted is not this one, but the one introduced by the same author that removed all the old url redirects

PR #776, opened 2016-06-14
Update protocols.markdown

jglover:
Appears that 'the' was accidentally used in place of 'to'.


josevalim:

PR #789, opened 2016-07-03
setup -> set up

kentor:
Should be the verb "set up" not the noun.


josevalim:

PR #791, opened 2016-07-05
How do we know this is working?

pablobm:
I was reading through this and felt that this section didn't explain sufficiently well how we can know that this change is working. I added some paragraphs to try a first approximation. You'll notice it's written like I don't know Elixir because... well, that.

Am I making sense? Would an addition along these lines be helpful to readers?

### DIFF
--- a/getting-started/mix-otp/supervisor-and-application.markdown
+++ b/getting-started/mix-otp/supervisor-and-application.markdown
@@ -203,7 +203,11 @@ iex> KV.Registry.lookup(KV.Registry, "shopping")
 {:ok, #PID<0.88.0>}
 ```
 
-Excellent!
+How do we know this is working? After all, we are creating the bucket and then looking it up; of course it should work, right? Well, remember that `KV.Registry.create/2` uses `GenServer.cast/3`, and therefore will return `:ok` regardless of whether the message finds its target or not. At that point, we don't know whether the supervisor and the server are up, and if the bucket was created. However, `KV.Registry.lookup/2` uses `GenServer.call/3`, and will block and wait for a response from the server. We do get a positive response, so we know all is up and running.
+
+For an experiment, try reimplementing `KV.Registry.create/2` to use `GenServer.call/3` instead, and momentarily disable the application callback. Run the code above on the console again, and you will see the creation step fail straightaway.
+
+Don't forget to bring the code back to normal before resuming this tutorial!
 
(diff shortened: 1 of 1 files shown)


getting-started/mix-otp/supervisor-and-application.markdown
```diff
@@ -203,7 +203,11 @@ iex> KV.Registry.lookup(KV.Registry, "shopping")
 {:ok, #PID<0.88.0>}
 ```
 
-Excellent!
+How do we know this is working? After all, we are creating the bucket and then looking it up; of course it should work, right? Well, remember that `KV.Registry.create/2` uses `GenServer.cast/3`, and therefore will return `:ok` regardless of whether the message finds its target or not. At that point, we don't know whether the supervisor and the server are up, and if the bucket was created. However, `KV.Registry.lookup/2` uses `GenServer.call/3`, and will block and wait for a response from the server. We do get a positive response, so we know all is up and running.
```
whatyouhide: I would replace "will return `:ok` regardless of whether the message finds it target or not" with something like "will return `:ok` regardless of whether the recipient of the message processes it or not" because Erlang guarantees that if you send a message locally it will be delivered to the recipient unless that recipient dies or something of that impact. Wdyt?

josevalim:
It looks great, thank you! 

getting-started/mix-otp/supervisor-and-application.markdown
josevalim: @whatyouhide cast does not guarantee it was delivered, therefore it does not guarantee it was processed. Since he is going with the weakest assumption (no delivery guarantee), I don't think we need to talk about processing guarantee. :)

whatyouhide: Yeah `cast` guarantees it's *sent*, but you always program on the assumption that if you send the message it *will* be delivered if nothing happens (like the recipient dying or something like that), even if with no guarantees on order/timeframe. But yeah, maybe the wording in the PR works well to explain the concept especially to newcomers, without going into too much detail :)

josevalim: cast does not guarantee it is sent because it will return `:ok` even if the process does not exist.

whatyouhide: And that is true! Right, sorry for all the noise then :)

PR #792, opened 2016-07-05
More specific explanation of keyword syntax for map

WojciechKo:


josevalim:
While your improvement is more complete, this is a getting started guide, so we don't need to cover all possible scenarios, only the most common ones. For this reason, I will close this for now. Thank you!  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

eksperimental:
Do we have this information in the elixir docs? I quickly read the docs for `Kernel.SpecialForms.%{}` and `Map` module, and it was not there.
We should definitely add it if not there yet

eksperimental:
I see it was **just** added by @whatyouhide in https://github.com/elixir-lang/elixir/pull/4947/files#diff-73ee63e805fc4f4b03d1581fb309de0bR21 to the documentation of Maps,
still i think we could add the example throwing `SyntaxError`

PR #794, opened 2016-07-07
Make closure example's captured variable explicit

feynmanliang:
The current example does not actually illustrate closure capture of variables in scope at function declaration.


josevalim:

whatyouhide:
Awesome catch, thanks @feynmanliang! 💟

PR #795, opened 2016-07-07
Inverted directory names in bucket_test.exs path

cgrande:
correct bucket_test.exs path


cgrande:
kv/test/kv looks odd as a path but seems intentional so I'll close

PR #796, opened 2016-07-07
Proposal to highlight indented code

riseshia:
Hi, I'm shia,
This is updated the indented, not syntax highlighted code to be applied syntax highlight.

meaning to say, indent will not be applied syntax highlight, so it look like blocked, strange T-T

how about give him to be more beautiful?
I can't find the reason why this snippet onl
y indented, not highlighted.

Thanks to read this. :)


josevalim:

PR #799, opened 2016-07-11
fixed non-compiling example

mreininger:
just a copy/paste fix


ericmj:
Nice catch. Thank you!

PR #802, opened 2016-07-13
Grammar fix: add missing "an"

laurenfackler:


whatyouhide:
I think there was no "an" here because we pass the supervisor as the *only* arguments to the bucket's `start_link/1` function... Does that make sense? Also, I am not a native English speaker so if the current wording is plain wrong let's fix it of course :)

joshfreemanIO:
The *a*/*an* indefinite article focuses upon the beginning sound of a word, not if the word is in a singular or plural form. So *an* would be more correct.  If we want to be more specific about the argument provided, then we would use the definite article *the*.

tl;dr: using *an*/*the* is fine.

whatyouhide:
Well thanks a lot then! 👍 💟 :)

PR #804, opened 2016-07-15
Updated Slack link in important-links.html

pikeview:


eksperimental:
we should be linking to https://elixir-lang.slack.com/ and let that site get the invitation, but it's not showing anything there.
any way that sign-in page can be updated?

josevalim:
I would prefer if the website contained instructions and we keep the instructions in setting up third party tools to a minimum.

pikeview:
Slack login page does not have "Request for invitation" link.  A few other communities, including erlang, have links to both slack login page and slack invite page.

stevedomin:
@eksperimental there is no way to customize the Slack sign-in page unfortunately.

@pikeview I think it's ok the way it is now, the elixir-slackin page is the best way onto the Elixir Slack imo. And it already contains a link to the Slack sign-in page if needed.

josevalim:
Thank you @stevedomin!

eksperimental:
well, given that the community is growing and growing, i think we should have a Community page, similar to https://www.erlang.org/community

We could explain there clearly to click on the invitation link and than follow the slack page.
We could also link to other important pages that are not in the right bar. such as the StackOverflown page. only linked in one place in this site, and that is the Introduction page.

WDYGT?

PR #805, opened 2016-07-15
simplify the pattern match example

bronson:
I spent a while trying to figure out why the pattern would match `42` but not `"!"`.  Turns out my browser doesn't show much difference between `[` and `{` so I was concentrating on the more obvious -- but irrelevant -- change.


bronson:
Dangit, I didn't update the result.  Wait a few minutes, going to rebase to fix.

bronson:
OK, ready now.  Confirmed the result with iex.

josevalim:

PR #807, opened 2016-07-19
GenStage. typos and small changes for clarity

jfis:


_posts/2016-07-14-announcing-genstage.markdown
```diff
@@ -67,21 +67,21 @@ File.stream!("path/to/some/file")
...
+First of all, we want to avoid moving data between processes as much as possible. Instead, we want to start multiple processes that perform the same computation in parallel. Not only that, if we are requiring developers to place `Stream.async` manually, it may lead to inefficient and error prone solutions.
 
 Although the solution above has many flaws, it has helped us ask the right questions:
 
   * If `Stream.async` is introducing new processes, how can we guarantee those processes are supervised?
 
-  * Since we are exchanging messages between processes, how can we do so with back-pressure? After all, if one process cannot process data as fast as it receives them, we want to slow down the processes sending the data so we guarantee the slowest process does not get overflown with messages
+  * Since we are exchanging messages between processes, how do we prevent a receiving process from being overwhelmed?
```
whatyouhide: I think a more verbose explanation here could still be easier to understand?

_posts/2016-07-14-announcing-genstage.markdown
```diff
@@ -319,7 +319,7 @@ File.stream!("path/to/some/file")
 |> Enum.to_list()
 ```
 
-While the above is helpful when working with large, possibly infinite collections, it still does not leverage concurrency. To address that, we are currently exploring a solution named [`GenStage.Flow`](https://hexdocs.pm/gen_stage/Experimental.GenStage.Flow.html), that allows us to express our computations similarly to streams, except they will run across multiple stages instead of a single process:
+While the above is helpful when working with large, possibly infinite, collections, it still does not leverage concurrency. To address that, we are currently exploring a solution named [`GenStage.Flow`](https://hexdocs.pm/gen_stage/Experimental.GenStage.Flow.html), that allows us to express our computations similarly to streams, except they will run across multiple stages instead of a single process:
```
whatyouhide: I think this comma hurts the readability a bit tbh 😕

_posts/2016-07-14-announcing-genstage.markdown
josevalim: Yes, can we expand a bit more on this?

_posts/2016-07-14-announcing-genstage.markdown
josevalim: Although it is the correct version. :P

whatyouhide: Yes, I think it is, but it just reads bad to me 😛 But your choice :)

jfis: ha ok then. how about "large or infinite collections" ?

_posts/2016-07-14-announcing-genstage.markdown
jfis: so i thought this was confusing to use process as a noun and a verb. i pared it down and listed only the question, as is hinted at by the "right questions" lead-in and the first question listed.  the explanation-- demand-driven back-pressure-- is what the rest of the article explains. 

but if it still needs a bit more, here's a stab at it:

    Since we are exchanging messages between processes, how do we prevent a process from receiving _too many_ messages? We need a mechanism for the receiving process to specify how much it can handle from the sending process. Of course, a sending process may not be able to supply the specified amount. With that in mind, we can say that the receiving process is merely specifying what it wants, or its _demand_. We can then call this mechanism "demand-driven back-pressure". Also, "sending process" and "receiving process" can become verbose rather quickly. We can call these "Producer" and "Consumer", respectively.

_posts/2016-07-14-announcing-genstage.markdown
josevalim: Let's go with "large or infinite collections", great suggestion.

_posts/2016-07-14-announcing-genstage.markdown
josevalim: What if we go with: "Since we are exchanging messages between processes, how do we prevent a process from receiving too many messages? We need a back-pressure mechanism that allows the receiving process to specify how much it can handle from the sending process."

?

_posts/2016-07-14-announcing-genstage.markdown
jfis: ok changed

_posts/2016-07-14-announcing-genstage.markdown
jfis: ok changed

josevalim:

PR #810, opened 2016-07-22
Make comments on ETS tests more consistent

semanticart:
This is a very small change for consistency's sake.

❤️ 

You can see the case differs from the other comment [here](https://github.com/elixir-lang/elixir-lang.github.com/pull/810/files#diff-8c5f16decd3fd0e493626da2abb4fe46R254)


eksperimental:
shouldn't it be  ":DOWN message"?

semanticart:
> shouldn't it be ":DOWN message"?

I'm happy to change both instances to that if there's a consensus.

josevalim:
It is fine as is. :) The code is the ultimate truth anyway!

josevalim:

PR #811, opened 2016-07-23
Reduce indent offset

gmile:
Noticed this while reading the docs for supervisor.


whatyouhide:
Great catch, thanks @gmile! 💟

gmile:
@whatyouhide no problem! 👍

PR #813, opened 2016-07-26
Remove bodyless function definition

eafif:


getting-started/mix-otp/docs-tests-and-with.markdown
```diff
@@ -289,7 +289,6 @@ The last step is to implement `KVServer.Command.run/1`, to run the parsed comman
 @doc """
 Runs the given command.
 """
-def run(command)
```
josevalim: This is on purpose. It will give us nice looking headers when generating the docs. :)

PR #815, opened 2016-07-28
Module#scripted-mode section: try to explain the diffrence between .ex and .exs files in a non-confusion way

sunboshan:
Per discussion in https://github.com/elixir-lang/elixir/issues/5073


josevalim:

PR #818, opened 2016-07-30
Revise sentence structure in 'Understanding Aliases'

1cph93:
Sentence should read as:
At this point, you may be wondering: what exactly is an Elixir alias and how is it represented?
or
At this point, you may be wondering what exactly an Elixir alias is and how it is represented.


josevalim:

PR #825, opened 2016-08-11

Ljzn:
An indentation error.


whatyouhide:
Thanks a lot @Ljzn! 💟

PR #827, opened 2016-08-12
fix broken Flow hexdocs links

ckampfe:
Hi folks,

Was reading the 7/14 Announcing GenStage post and noticed a few broken links that were pointing to the hexdocs for `Experimental.GenStage.Flow`, which has since been renamed to `Experimental.Flow`. I left the actual names (`GenStage.Flow`) alone to reflect what the module was called at time of writing.


josevalim:

PR #831, opened 2016-08-18
Update case-cond-and-if.markdown

javs-perez:
Just a minor fix.


josevalim:

PR #832, opened 2016-08-18
Point to point to the "Operators" page in the Elixir docs

whatyouhide:
The logical next step after https://github.com/elixir-lang/elixir/pull/5142 :) \cc @josevalim, I opened a PR just to check that it's ok to point to `/docs/master` instead of `/docs/stable` since I assume the new page won't be in `/docs/stable` yet, right?


josevalim:
:shipit: ! And yeah, master is fine for now!

Issue #834, opened 2016-08-22
Clearly mark chapters as being part of the "Mix and OTP" guide


sotte:
The chapters of the "Mix and OTP" guide are, no surprise here, part of the "Mix and OTP" guide and often depend on the previous chapters. When a google result brings you to one of the chapters this is not apparent though.

I would add a little note to the beginning of all chapters in the "Mix and OTP" guide, something like this

```
Note that this chapter is part of the "Mix and OTP" guide and depends on the previous chapters.
Maybe read the [TODO add link](previous chapter) or start reading from the [http://elixir-lang.org/getting-started/mix-otp/introduction-to-mix.html](first chapter) of the guide.
```

If you're interested in this I would create a pull request.

whatyouhide:
This is not a bad idea IMO, and I lean on being 👍 on this (or something similar to this). Let's hear what the others think :)

josevalim:
Sounds good to me!

sotte:
OK, I'll put a pull request together tomorrow.

PR #836, opened 2016-08-24
Mark "Mix & OTP" chapters as part of the guide

sotte:
In the beginning of every chapter reference the previous chapter and the first
chapter of the "OTP & Mix" guide.

See #834

I can change the wording if required.

### DIFF
--- a/getting-started/mix-otp/agent.markdown
+++ b/getting-started/mix-otp/agent.markdown
@@ -7,6 +7,8 @@ title: Agent
 
 {% include toc.html %}
 
+{% include mix-otp-preface.html %}
+
 In this chapter, we will create a module named `KV.Bucket`. This module will be responsible for storing our key-value entries in a way it can be read and modified by other processes.
 
 If you have skipped the Getting Started guide or if you have read it long ago, be sure to re-read the chapter about [Processes](/getting-started/processes.html). We will use it as a starting point.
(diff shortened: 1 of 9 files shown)


getting-started/mix-otp/agent.markdown
```diff
@@ -7,6 +7,8 @@ title: Agent
 
 {% include toc.html %}
 
+Note that this chapter is part of the *Mix and OTP* guide and it depends on the previous chapters. Maybe read the [previous chapter](/getting-started/mix-otp/introduction-to-mix.html).
```
whatyouhide: I would say 

> Note that this chapter is part of the *Mix and OTP* guide and it depends on previous chapters in this guide. Here's [the previous chapter](...)

Wdyt?

sotte: Fine with me!

I'll wait for more feedback and update the PR later.

josevalim:
I would prefer to go with something like:

> This chapter is part of the *Mix and OTP* guide and it depends on previous chapters in this guide. For more information, [read the guide introduction](...) or check out the chapter index in the sidebar.

If we keep it generic, we can even put it in a include and use it in all chapters.

Thoughts?

whatyouhide:
@josevalim yeah, I was leaning on that too tbh. Let's go with a generic link to the whole guide 👍

sotte:
I'm using an `include` now. Didn't touch the intro.

josevalim:

PR #838, opened 2016-08-26
Updated the page on sigils

Qqwy:
The page about sigils is updated, specifically the section about defining your own sigils:

- clarified that only single-letters can be used.
- clarified that lowercase letters allow interpolation, uppercase does not.
- clarified that the unescaping of special characters needs to happen manually in your sigil implementation.

I made this update because I was confused by this page; see https://github.com/elixir-lang/elixir/issues/5153 .

### DIFF
--- a/getting-started/sigils.markdown
+++ b/getting-started/sigils.markdown
@@ -176,7 +176,9 @@ iex> h sigil_r
 ...
 ```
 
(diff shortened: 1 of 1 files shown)


josevalim:
Thank you for the PR although I would prefer to not merge it. This is a getting started guide. It is not meant to be a complete reference to how the language works, specially an advanced topic like sigils. We should keep it simple and the documentation is there for those who want to go deeper.

Qqwy:
All right, but can we at least add a link to the appropriate section in the documentation?

PR #847, opened 2016-09-19
Rename variable in `unless` example

telnicky:
The name `options` for the second argument didn't seem to make much sense given that the implementation is `unless`. What do you think about renaming the variable to `expression`?

### DIFF
--- a/getting-started/meta/macros.markdown
+++ b/getting-started/meta/macros.markdown
@@ -23,11 +23,11 @@ In order to better understand how macros work, let's create a new module where w
 
 ```elixir
 defmodule Unless do
-  def fun_unless(clause, expression) do
+  def fun_unless(clause, do: expression) do
     if(!clause, do: expression)
   end
 
-  defmacro macro_unless(clause, expression) do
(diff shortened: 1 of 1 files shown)


whatyouhide:
We usually call this `options` anyways because really it's a keyword list, and the "expression" is in the `:do` keyword. I actually think the example linked by @eksperimental is wrong, it should be `macro_unless(clause, do: expression)`? :)

telnicky:
Oh interesting, you are right a more correct implementation would contain the `:do` keyword. It appears the current example is still correct if you only supply a single expression. For the purpose of teaching macros, do we think it'd be better to keep it simple or adhere closer to the actual implementation?

whatyouhide:
@telnicky I would prefer using `do: expression` in the `unless` example as well, after all I don't think it's substantially more complicated. Wdyt?

telnicky:
I agree, especially after taking a pass at adding the `:do`'s. What do you all think now?

josevalim:

eksperimental:
Great!

PR #848, opened 2016-09-21
Correct types in sorting order

eksperimental:
based on 
https://github.com/elixir-lang/elixir/pull/5238/files#diff-50948d471af46f8f79bf8d175641fa42R67
by @fboyer

### DIFF
--- a/getting-started/basic-operators.markdown
+++ b/getting-started/basic-operators.markdown
@@ -107,7 +107,7 @@ true
 
 The reason we can compare different data types is pragmatism. Sorting algorithms don't need to worry about different data types in order to sort. The overall sorting order is defined below:
 
-    number < atom < reference < functions < port < pid < tuple < map < list < bitstring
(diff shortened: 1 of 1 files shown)


eksperimental:
please let me know if you need the commits to be squashed

whatyouhide:
Do we still need this in the guides @josevalim, or would it be wise to reword this chapter so that the heavy stuff is relegated to the Elixir docs?

josevalim:
@whatyouhide I would still keep the ordering and say: "You don't need to memorize this ordering, it is enough to know an ordering exists. For more information, check the reference page on Operators."

Btw, what if we start calling the pages in the docs "Reference pages"? Since their goal is not to actually guide but provide a complete reference on a given topic?

whatyouhide:
Yeah "Reference page" is a good description of what they are. So let's keep the ordering and write what you said, but let's merge this one like it is right? I can reword directly on master.

josevalim:

PR #849, opened 2016-09-22
Document how to use strong crypto seeds with rand

dkarter:
Based on @josevalim's recommendation here:
https://groups.google.com/forum/#!topic/elixir-lang-core/-sKe8sA3cNI

And this blog post:
https://hashrocket.com/blog/posts/the-adventures-of-generating-random-numbers-in-erlang-and-elixir

Edits welcome :)

### DIFF
--- a/getting-started/erlang-libraries.markdown
+++ b/getting-started/erlang-libraries.markdown
@@ -172,6 +172,21 @@ iex> :rand.uniform(6)
 6
 ```
 
+`rand`'s documentation states that `rand`'s default seed is not cryptographically
+strong. It suggests using the `crypto` library for seeding `rand` with a
+cryptographically strong seed when needed.
+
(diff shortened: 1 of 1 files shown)


josevalim:
Hi @dkarter!

Sorry, I believe I was not clear enough. I proposed to either link to your blog post from here or add such examples to Erlang documentation. However, I am also ok with this current pull request, but I thought I would make myself clearer and get your feedback before taking a decision.

Thoughts?

fishcakez:
@dkarter I think you have misunderstood the `:rand` documentation. The `:rand` pseudo random number generator is not cryptographically strong. Unfortunately seeding with `:crypto` does not make it so. I believe for your intensions all numbers must be generated using `:crypto.strong_random_bytes`.

PR #851, opened 2016-09-25
Add unreachable line back plus a note

gusaaaaa:
On a second thought, if we remove the line we should also remove line 180, and probably need to explain why we added `loop_acceptor(socket)` in line 190.

I recommend adding `loop_acceptor(socket)` back after line 53 and add a small note explaining why the line is there.

Maybe the note's wording does not match the writing style, so please fill free to change it as you wish.

### DIFF
--- a/getting-started/mix-otp/task-and-gen-tcp.markdown
+++ b/getting-started/mix-otp/task-and-gen-tcp.markdown
@@ -51,6 +51,7 @@ defp serve(socket) do
   |> write_line(socket)
 
   serve(socket)
+  loop_acceptor(socket)
 end
 
 defp read_line(socket) do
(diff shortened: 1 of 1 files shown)


josevalim:
@gusaaaaa maybe it would be better to not add it back and add it later on and explain why we added it later on indeed? that's preferrable that adding it now and saying we will need it later, no? :)

gusaaaaa:
@josevalim I'm hesitant. Removing the tail call may make some people think why the function is ever called `loop_acceptor/1`, if it is not a loop. We could explain that it is not a loop for now, but it will be latter... However, the explanation will get even more complex.

For didactic purposes I would leave the tail call right there, even if it the code temporary flawed in some way. Leaving it there may even have an extra benefit: making people who didn't realize about it, question why the code is unreachable.

josevalim:

gusaaaaa:
@josevalim Oh, sh*t! I added the call to `loop_acceptor(socket)` in line 54 by mistake. I should've added it to line 46. I'm amending this on PR https://github.com/elixir-lang/elixir-lang.github.com/pull/852. Sorry for this.

PR #852, opened 2016-09-25
Fix call to loop_acceptor

gusaaaaa:
Amends changes made in PR https://github.com/elixir-lang/elixir-lang.github.com/pull/851. The tail call to `loop_acceptor/1` should go at the end of `loop_acceptor/1`.


josevalim:

PR #854, opened 2016-09-28
Set larger text size

vladdu:
This may affect everything, because it's a setting on `body`. It looks better to me, but I'll leave the decision to you :-)

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -33,7 +33,7 @@ html {
 }
 body {
   margin: 0;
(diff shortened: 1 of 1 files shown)


josevalim:
Thank you. I think we can go at least with 0.9em and move the menu to 1.0em. Increasing more than that for now causes a large contrast since we still need to fix other parts of the codebase. :) I will push it soon.

PR #856, opened 2016-09-30
Update supervisor-and-application.markdown

chareice:
context is a map,not a function.

### DIFF
--- a/getting-started/mix-otp/supervisor-and-application.markdown
+++ b/getting-started/mix-otp/supervisor-and-application.markdown
@@ -70,12 +70,12 @@ We also need to update our tests to give a name when starting the registry. Repl
 
 ```elixir
 setup context do
-  {:ok, registry} = KV.Registry.start_link(context.test)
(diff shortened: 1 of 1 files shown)


getting-started/mix-otp/supervisor-and-application.markdown
```diff
@@ -70,12 +70,12 @@ We also need to update our tests to give a name when starting the registry. Repl
 
 ```elixir
 setup context do
-  {:ok, registry} = KV.Registry.start_link(context.test)
+  {:ok, registry} = KV.Registry.start_link(context[:test])
```
josevalim: `context` is a map and you can access fields of a map using the `map.field` syntax. :)

josevalim:
Thank you but the current code is also correct as is. :)

eksperimental:
Shouldn't this be mentioned in `Kernel.SpecialForms` `./2` documentation?

josevalim:
@eksperimental surely :D

chareice:
@josevalim Thanks，since i'm a beginner in elixir😅

PR #866, opened 2016-10-18
Update description for adding run method

kdelchev:
IMO current version of the tutorial does not make clear a new method must be added. I stumbled upon this and replaced 
```
  def run({:create, bucket}) do
    KV.Registry.create(KV.Registry, bucket)
    {:ok, "OK\r\n"}
  end
```

with

```
  def run({:create, bucket}, pid) do
    KV.Registry.create(pid, bucket)
    {:ok, "OK\r\n"}
  end
```
instead adding another "version" of the method. This confusion could come with OOP way of thinking. I am not a native speaker, so the wording could be changed.


josevalim:
Hi @kdelchev. The idea is *not* to add a new clause nor change the existing ones. That is just an example of how we could write those but later on we decide to go with an integration test. I will update the chapter to make it clearer, thank you!

PR #871, opened 2016-11-10
Three small edits to add clarity and conciseness

rawreef:
An attempt to simplify and add clarity to the text. (I'm an Elixir newbie, but I took a gamble on changing "the sides can't match" to "the sides can't be matched". Based on what I've understood from Dave Thomas's book (the first few pages), I think my version is clearer. Of course I may be wrong!)

One question, is making a pull request with these small edits a useful thing to do, or should I just do it when I see a more serious error in the text? I want to be seen as helpful rather than pedantic to the Elixir community. :-)

### DIFF
--- a/getting-started/pattern-matching.markdown
+++ b/getting-started/pattern-matching.markdown
@@ -53,7 +53,7 @@ iex> b
 "world"
 ```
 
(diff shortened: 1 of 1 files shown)


whatyouhide:
Great job @rawreef :)

josevalim:

PR #872, opened 2016-11-13
Amend explanation of `with` construct in mix-otp guide

lulessa:
When introducing `with`, explain briefly how the construct can replace
nested `case` clauses.

Additionally, fix a typo referring to KVServer.

### DIFF
--- a/getting-started/mix-otp/docs-tests-and-with.markdown
+++ b/getting-started/mix-otp/docs-tests-and-with.markdown
@@ -265,7 +265,7 @@ This means our implementation is going in the correct direction, but it doesn't 
 
 The previous implementation used pipelines which made the logic straight-forward to follow. However, now that we need to handle different error codes along the way, our server logic is nested inside many `case` calls.
 
-Thankfully, Elixir v1.2 introduced a construct called `with` which allows to simplify code like above. Let's rewrite the `serve/1` function to use it:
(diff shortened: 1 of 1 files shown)


getting-started/mix-otp/docs-tests-and-with.markdown
```diff
@@ -265,7 +265,7 @@ This means our implementation is going in the correct direction, but it doesn't
 
 The previous implementation used pipelines which made the logic straight-forward to follow. However, now that we need to handle different error codes along the way, our server logic is nested inside many `case` calls.
 
-Thankfully, Elixir v1.2 introduced a construct called `with` which allows to simplify code like above. Let's rewrite the `serve/1` function to use it:
+Thankfully, Elixir v1.2 introduced the `with` construct, which allows you to simplify code like the above, replacing nested `case` calls with a chain of matching clauses. Let's rewrite the `serve/1` function to use `with`:
```
whatyouhide: "simplify code like the above" -> "simplify code like the one above"? Not sure 😕

josevalim:

getting-started/mix-otp/docs-tests-and-with.markdown
lulessa: Now I'm thinking we don't need to mention the "code above," at all. There's enough context and it's clear that the target of our refactoring will be nested `case` calls. Maybe on my next PR I'll change it to "allows you to replace nested `case` calls with a chain of matching clauses."

PR #875, opened 2016-11-14
Assorted edits

fxn:
This patch has assorted edits.

**Oxford comma:** We commented a few days ago with @josevalim that the docs do not use the Oxford comma consistently. It is one of those style details that are worth having uniform across the docs and you need to pick one option and stick to it. Going with the Oxford comma seems to be the safest choice because [a majority of style guides](https://en.wikipedia.org/wiki/Serial_comma) of American English mandate its use.

**Error message of `hd []`:** Completes the output to match the real one.

**`size` vs `length`:** There is a mnemonic to easily remember which of `size` or `length` mean "linear".

**First `iex` session:** The first `iex` session the reader finds has one partial banner. Since this is the first time a session like this is shown I believe readers get more confidence if what they see in the console matches as closely as possible. So I have added the Erlang banner and shell prompts counter. I guess readers will understand the version numbers and minor details may differ.

From then on, it is understandable that `iex` sessions are shown stripped down.

Also, since this is the first session, the patch stresses how to get out of the shell, since the usual `Ctrl-D` and friends do not work, and the reader may have overlooked the message in the banner.

**WhatsApp link:** The HTTP link of WhatsApp redirects to HTTPS, easier to point to that one.

### DIFF
--- a/getting-started/basic-types.markdown
+++ b/getting-started/basic-types.markdown
@@ -46,7 +46,7 @@ iex> rem 10, 3
 
 Notice that parentheses are not required in order to invoke a function.
 
-Elixir also supports shortcut notations for entering binary, octal and hexadecimal numbers:
+Elixir also supports shortcut notations for entering binary, octal, and hexadecimal numbers:
 
 ```iex
 iex> 0b1010
@@ -99,7 +99,7 @@ iex> is_boolean(1)
 false
 ```
 
-You can also use `is_integer/1`, `is_float/1` or `is_number/1` to check, respectively, if an argument is an integer, a float or either.
+You can also use `is_integer/1`, `is_float/1` or `is_number/1` to check, respectively, if an argument is an integer, a float, or either.
 
 > Note: At any moment you can type `h` in the shell to print information on how to use the shell. The `h` helper can also be used to access documentation for any function. For example, typing `h is_integer/1` is going to print the documentation for the `is_integer/1` function. It also works with operators and other constructs (try `h ==/2`).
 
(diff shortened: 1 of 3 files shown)


getting-started/introduction.markdown
```diff
@@ -34,14 +34,17 @@ For now, let's start by running `iex` (or `iex.bat` if you are on Windows) which
 Open up `iex` and type the following expressions:
 
 ```iex
-Interactive Elixir - press Ctrl+C to exit (type h() ENTER for help)
+Erlang/OTP 19 [erts-8.1] [source] [64-bit] [smp:4:4] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]
```
fxn: @eksperimental you'd like the banners to match the oldest versions? do you know how would they look like?

getting-started/basic-types.markdown
```diff
@@ -270,6 +270,7 @@ Getting the head or the tail of an empty list is an error:
 ```iex
 iex> hd []
 ** (ArgumentError) argument error
+    :erlang.hd([])
```
whatyouhide: I am torn on this, as this could be considered a "stacktrace" and we usually don't show stacktraces when exceptions happen in `iex`. 😕

getting-started/introduction.markdown
whatyouhide: Instead of tying ourselves to a version, we can use `19` and say "your prompt will look something like this, depending on the version" like many books do.

whatyouhide:
👏 on the Oxford comma! 💟

fxn:
@whatyouhide Roger the convention re backtraces, I have removed it.

getting-started/introduction.markdown
fxn: That sounds like a good compromise. The reader understands you pasted _some_ session, and with time details differ.

Alternatively, this shell could be stripped down like the rest. I had the feeling that it was half way, neither simplified, nor complete.

fxn:
@eksperimental @whatyouhide have pushed one possible way to address it, what do you think?

getting-started/introduction.markdown
eksperimental: @fxn if you are still interested, this is what 18.0 looks like
> Erlang/OTP 18 [erts-7.3] [source] [smp:4:4] [async-threads:10] [kernel-poll:false]

getting-started/introduction.markdown
```diff
@@ -34,14 +34,17 @@ For now, let's start by running `iex` (or `iex.bat` if you are on Windows) which
...
-iex> 40 + 2
+Interactive Elixir (1.3.4) - press Ctrl+C to exit (type h() ENTER for help)
+iex(1)> 40 + 2
 42
-iex> "hello" <> " world"
+iex(2)> "hello" <> " world"
 "hello world"
 ```
 
+Please note that some details like version numbers may differ a bit in your session, that's not important. From now own `iex` sessions will be stripped down to focus on the code. To exit `iex` press `Ctrl+C` twice.
```
eksperimental: "from now ON".

eksperimental:
@fxn I agree with the notice.

fxn:
@eksperimental "on" fixed, thanks!

Regarding the Erlang banner... I'll leave the decision to you guys, whatever you believe is better is going to be fine with me (and I'll revise the patch as needed).

whatyouhide:
Great work @fxn, thanks! 💟

fxn:
@whatyouhide I am going through the page about `case` and friends and saw

```
iex> hd(1)
** (ArgumentError) argument error
    :erlang.hd(1)
```

Should `:erlang.hd(1)` be removed there? Or should we backtrack here? Or none of them? :smile:

whatyouhide:
@fxn I would remove the stacktrace there as well :)

fxn:
@whatyouhide awesome, PR coming soon.

PR #876, opened 2016-11-15
More assorted edits

fxn:
More edits:

* Oxford comma in a few places.
* Removes `:erlang.hd(1)` according to https://github.com/elixir-lang/elixir-lang.github.com/pull/875#issuecomment-260584786.
* A missing space.
* The patch changes **keyword list** to _keyword list_ because in previous sections new technical terms go in italics, like _pin operator_. Also, using italics when vocabulary is first introduced is a pretty standard convention (so no need to change the previous ones to bold face, going with italics is fine).
* I have reworded the explanation about why does `is_number(if true)` fail. The current description says Elixir tries to invoke `is_number/1`. I don't know the internals, but from a user's perspective I think the idea is that Elixir is going to look for a function with the arity parsed at the call point, so looking for `is_number/2`, which does not exist. Done. This is different from calling `is_number/1` and fail. Please, if this mental model is wrong just tell me and I'll roll the change back.

### DIFF
--- a/getting-started/basic-operators.markdown
+++ b/getting-started/basic-operators.markdown
@@ -78,7 +78,7 @@ true
 
 As a rule of thumb, use `and`, `or` and `not` when you are expecting booleans. If any of the arguments are non-boolean, use `&&`, `||` and `!`.
 
-Elixir also provides `==`, `!=`, `===`, `!==`, `<=`, `>=`, `<` and `>` as comparison operators:
+Elixir also provides `==`, `!=`, `===`, `!==`, `<=`, `>=`, `<`, and `>` as comparison operators:
 
 ```iex
 iex> 1 == 1
@@ -113,4 +113,4 @@ You don't actually need to memorize this ordering, it's enough to know that this
 
 For reference information about operators (and ordering), check the [reference page on operators](/docs/master/elixir/operators.html).
 
-In the next chapter, we are going to discuss some basic functions, data type conversions and a bit of control-flow.
+In the next chapter, we are going to discuss some basic functions, data type conversions, and a bit of control flow.
--- a/getting-started/case-cond-and-if.markdown
+++ b/getting-started/case-cond-and-if.markdown
@@ -1,13 +1,13 @@
 ---
 layout: getting-started
-title: case, cond and if
+title: case, cond, and if
 ---
 
 # {{ page.title }}
 
 {% include toc.html %}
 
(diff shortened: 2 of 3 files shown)


getting-started/case-cond-and-if.markdown
```diff
@@ -224,14 +223,14 @@ iex> if nil do
...
-At this point, we have learned four control structures: `case`, `cond`, `if` and `unless`, and they were all wrapped in `do/end` blocks. It happens we could also write `if` as follows:
+At this point, we have learned four control structures: `case`, `cond`, `if`, and `unless`, and they were all wrapped in `do/end` blocks. It happens we could also write `if` as follows:
 
 ```iex
 iex> if true, do: 1 + 2
 3
 ```
 
-Notice how the example above has a comma between `true` and `do:`, that's because it is using Elixir's regular syntax where each argument is separated by comma. We say this syntax is using **keyword lists**. We can pass `else` using keywords too:
+Notice how the example above has a comma between `true` and `do:`, that's because it is using Elixir's regular syntax where each argument is separated by comma. We say this syntax is using _keyword lists_. We can pass `else` using keywords too:
```
whatyouhide: I'm not sure about the rest of the guide, but in the Elixir docs we use `*italic*` instead of `_italic_`. Can we do the same here? :)

getting-started/basic-operators.markdown
```diff
@@ -113,4 +113,4 @@ You don't actually need to memorize this ordering, it's enough to know that this
 
 For reference information about operators (and ordering), check the [reference page on operators](/docs/master/elixir/operators.html).
 
-In the next chapter, we are going to discuss some basic functions, data type conversions and a bit of control-flow.
+In the next chapter, we are going to discuss some basic functions, data type conversions, and a bit of control-flow.
```
whatyouhide: Since we're here and we modified this line, wdyt if we remove the hyphen in "control-flow"? I think it's unnecessary as far as my English-fu goes.

fxn:
@whatyouhide cools it's pushed, and agree wrt the hyphen in "control-flow". I have grepped the project and have edited another occurrence (one remains in a blog post, not sure if that one should be edited.)

whatyouhide:
I am for removing the one in the blog post as well but let's do this: I'll approve this now and let @josevalim take the decision about the blog post. :)

josevalim:

PR #878, opened 2016-11-22
Revises Unicode terminology

fxn:
This patch revises the usage of the term "code point" in the section of the Getting Started guide about strings.

In Unicode terminology, a _code point_ [is an integer](http://unicode.org/glossary/#code_point), and the glyph is a _character_, so the wording needs to be precise in that sense: Say "code point" to refer to the code, and say "character" when you are talking about its visual representation.

Similarly, make crystal clear to the reader that a char list is a list of integers that satisfy some ad-hoc condition, and that single-quoted literals allow you to create such lists.


josevalim:

PR #880, opened 2016-11-24
Update install section about Fedora and dnf

jandudulski:
Since Fedora 22 they switched from yum to dnf tool.

More: https://fedoraproject.org/wiki/F22_release_announcement#Faster_and_better_dependency_management_with_DNF


josevalim:

PR #882, opened 2016-11-29
Update installation instructions for latest ubuntu version

tsubery:
This installation procedure works for ubuntu 16.04

### DIFF
--- a/install.markdown
+++ b/install.markdown
@@ -44,7 +44,7 @@ If your distribution contains an old Elixir/Erlang version, see the sections bel
   * FreeBSD
     * From ports: `cd /usr/ports/lang/elixir && make install clean`
     * From pkg: `pkg install elixir`
-  * Ubuntu 12.04 and 14.04 / Debian 7
+  * Ubuntu 12.04/14.04/16.04 or Debian 7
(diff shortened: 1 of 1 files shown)


josevalim:

install.markdown
```diff
@@ -44,7 +44,7 @@ If your distribution contains an old Elixir/Erlang version, see the sections bel
   * FreeBSD
     * From ports: `cd /usr/ports/lang/elixir && make install clean`
     * From pkg: `pkg install elixir`
-  * Ubuntu 12.04 and 14.04 / Debian 7
+  * Ubuntu 12.04/14.04/16.04 or Debian 7
```
eksperimental: this will most likely work for 12.10, 14.10 and 16.10,
so i'm wondering if we should just mention from 12.04 onwards...

tsubery:
13 &15 have passed End of life so they won't likely work. We could somehow
phrase this exception

eksperimental:
sorry, but why wouldn't they work?

tsubery:
I actually tried spawning a 13.04 vm and test it out before submitting the pull request. After few minutes of trying to figure out why apt-get can't find the necessary dependencies I realized it was long since this version passed End of Life. I can imagine other users going through the same process if it specifically claimed to be supported.

eksperimental:
well, it does not necessarily mean that software stops working after its life support's expired. but it will make our lives easier if we just limit ourselves to give advice to whatever is still being supported.

PR #884, opened 2016-12-02
Updates the output of an example

fxn:
The "concat.ex:7" bit goes now below the warning, indented as backtrace:

```
warning: this clause cannot match because a previous clause at line 2 always matches
  concat.ex:7
```


josevalim:

PR #888, opened 2016-12-08
Update io-and-the-file-system.markdown

lambrospetrou:
I think there is a logic error in the intro sentence explaining why the chapter is earlier rather than later in the guide.


josevalim:
No, it is correct. IO is usually in introductory chapters and here it is past the middle. :)

PR #889, opened 2016-12-10
Fixes typo in Protocols section of getting started

jmnsf:
Just started learning today, but fairly sure that should be a Module and not a Protocol.


whatyouhide:
That was my bad, sorry! :( Good catch @jmnsf! 💟

PR #890, opened 2016-12-15
Add short paragraph about arity in getting started/basic types chapter

jessica-jordan:
Just started with the "Getting started" section and coming from a JavaScript background got quite hung up on the sudden introduction of arities. Let me know if you have any other suggestions and I'll update my PR!

### DIFF
--- a/getting-started/basic-types.markdown
+++ b/getting-started/basic-types.markdown
@@ -77,6 +77,10 @@ iex> trunc(3.58)
 3
 ```
 
+## Identifying function
+
+Functions in Elixir are identified by both their name and their arity. The arity of a function describes the number of arguments which the function takes. From this point on we will use both the function name and its arity to decribe functions throughout the documentation. `round/1` identifies the function which is named `round` and takes 1 argument, whereas `round/2` identifies a different (nonexistent) function with the same name but with an arity of only `2`.
+
 ## Booleans
 
 Elixir supports `true` and `false` as booleans:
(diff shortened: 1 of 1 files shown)


josevalim:
Thank you for the PR @jessica-jordan!

The PR is removing arities from all functions. I would prefer to not remove them and instead try to make it clearer why there us such a thing as arity, as well as their notation. Is there something else we can do to improve it?

jessica-jordan:
Hi @josevalim Thanks for your quick feedback! In fact, I removed all the arities from the functions only until it is explained in more detail with the newly added paragraph in the anonymous functions section. After that point arities are still used to describe functions. I just found it confusing to always see the full function description `function_name/arity_num` before they are explained beyond a side note. 
But one idea that comes to my mind is to start a short paragraph on function descriptions at the very top of the chapter, explaining what the syntax means and then go on with the other data types (integers, floats, booleans) after that and also describing anonymous functions in greater detail later on.

Looking forward to your and others' feedback.

whatyouhide:
> But one idea that comes to my mind is to start a short paragraph on function descriptions at the very top of the chapter, explaining what the syntax means and then go on with the other data types (integers, floats, booleans) after that and also describing anonymous functions in greater detail later on.

I think I like this idea better. I would prefer to educate readers about `function/arity` as soon as possible since it's all over the place in Erlang/Elixir :)

josevalim:
@jessica-jordan what do you think about having an explanation about this syntax before the Booleans section? Just after basic arithmetic? In basic arithmetic we already use `round` and `trunc` without the proper notation (on purpose) so after that may be a good point to introduce the concept?

jessica-jordan:
@josevalim @whatyouhide Sounds great to me, I'll update my PR in a bit

jessica-jordan:
Updated the PR now and added a short intro into function identification just after basic arithmetics. Let me know what you think

josevalim:

PR #891, opened 2016-12-18
Edits in processes.markdown

fxn:
The main contribution in this patch is explaining that `send/2` does not block. This is an introduction to processes, and the text says that `receive` waits. People new to the language would naturally wonder at that point if senders wait for their message to be consumed by the recipient.

The rest are details here and there seen on passing.


josevalim:

PR #892, opened 2016-12-19
Update error message for unrequired modules

JagdeepSingh:
Elixir (1.3.4) gives the updated error message for un`require`d modules.

### DIFF
--- a/getting-started/alias-require-and-import.markdown
+++ b/getting-started/alias-require-and-import.markdown
@@ -82,7 +82,11 @@ Macros are chunks of code that are executed and expanded at compilation time. Th
 
 ```iex
 iex> Integer.is_odd(3)
-** (CompileError) iex:1: you must require Integer before invoking the macro Integer.is_odd/1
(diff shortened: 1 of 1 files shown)


josevalim:
This week we are releasing Elixir v1.4 which says:

```iex
iex(1)> Integer.is_odd(1)
** (UndefinedFunctionError) function Integer.is_odd/1 is undefined or private but Integer defines a macro with the same name and arity. Be sure to require Integer if you intend to invoke this macro
    (elixir) Integer.is_odd(1)
```

Maybe we should use that instead?

Thank you!

JagdeepSingh:
Sounds good.

josevalim:
Awesome, can you please amend your PR then? Then it should be good to merge!

JagdeepSingh:
Ok. i will wait for the release and fix it then.

josevalim:
v1.4.0 is out. :)

JagdeepSingh:
Not needed now. The error message is correct on v1.4.0.

PR #894, opened 2016-12-20
Implement protocol for BitString to make it work for strings

JagdeepSingh:
Implementing the protocol for `Binary` does not work with strings. Instead we need to implement it for `BitString` in 1.3.4.

I guess this should be fixed (hopefully soon) as it is related to the code, not just the documentation or error messages. Please correct me, if i got it wrong.


josevalim:

whatyouhide:
Great catch! 💟

PR #895, opened 2016-12-20
Edits to genserver getting started guide for clarity.

russolsen:
As I read through the genserver getting started guide I found myself mentally editing and rearranging the prose a little bit. In particular, as a complete beginner, I was confused as to why the create operation was implemented as a cast, only to discovered towards the bottom of the document that it was only for illustrative purposes.

In any case, here are my suggestions.

### DIFF
--- a/getting-started/mix-otp/genserver.markdown
+++ b/getting-started/mix-otp/genserver.markdown
@@ -46,7 +46,9 @@ We will use a [GenServer](https://hexdocs.pm/elixir/GenServer.html) to create a 
 
 ## Our first GenServer
 
-A GenServer is implemented in two parts: the client API and the server callbacks. You can either combine both parts into a single module or you can separate them into a client module and a server module. The client and server run in separate processes, with the client passing messages back and forth to the server as its functions are called. Here we'll use a single module for both the server callbacks and client API. Create a new file at `lib/kv/registry.ex` with the following contents:
+A GenServer is implemented in two parts: the client API and the server callbacks. You can either combine both parts into a single module or you can separate them into a client module and a server module. The client and server run in separate processes, with the client passing messages back and forth to the server as its functions are called. Here we'll use a single module for both the server callbacks and the client API.
+
+Create a new file at `lib/kv/registry.ex` with the following contents:
 
 ```elixir
 defmodule KV.Registry do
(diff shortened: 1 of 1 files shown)


getting-started/mix-otp/genserver.markdown
```diff
@@ -42,11 +42,13 @@ Instead of abusing the name registry facility, we will create our own *registry
 
 The registry needs to guarantee that the dictionary is always up to date. For example, if one of the bucket processes crashes due to a bug, the registry must clean up the dictionary in order to avoid serving stale entries. In Elixir, we describe this by saying that the registry needs to *monitor* each bucket.
 
-We will use a [GenServer](https://hexdocs.pm/elixir/GenServer.html) to create a registry process that can monitor the bucket processes. GenServers are the go-to abstraction for building generic servers in both Elixir and  <abbr title="Open Telecom Platform">OTP</abbr>.
+We will use a [GenServer](https://hexdocs.pm/elixir/GenServer.html) to create a registry process that can monitor the bucket processes. GenServer provides an industrial strength framework for building servers in both Elixir and  <abbr title="Open Telecom Platform">OTP</abbr>.
```
josevalim: Maybe I am nitpicking but wouldn't "abstraction" be a better description than "framework" given the expectation most developers have regarding what a "framework" is?

josevalim:
I have dropped a single comment, everything else looks great. Thank you!

getting-started/mix-otp/genserver.markdown
russolsen: The reason I changed it from "abstraction" to, well, something else, is that I wanted to get across that GenServer actually provides the programmer with real world help in building applications - when you use GenServer you aren't just adopting an idea, you are pulling in some code. I picked "framework" because GenServer is driving - in a sense it provides the "main program" for the server process and then calls back into your functions. This in my mind is more or less the definition of a framework.

Having said that I realize I'm a complete beginner here and I'm fine with whatever you do.

eksperimental: We always refer t them as abstractions, it will be misleading to call them different just here.

josevalim:
Thank you @russolsen! I will merge this as is and settle on framework or abstraction or none after merging.  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #900, opened 2016-12-26
Edits in the "alias, require, and import" chapter

fxn:
Besides the details like the Oxford comma and such, the main edit in this patch rewords a bit the introduction of `alias`.

The original example has a complete module definition

```elixir
defmodule Math do
  alias Math.List, as: List
end
```

and the text starts saying

> From now on, any reference to `List` will automatically expand to `Math.List`.

If you look at the two, you'll see "From now on" is not as clear as it could be, because there is not even a place in which the alias is effective. The text introduces `alias` with an example in which the directive has no effect. Moreover, at this point the reader does not know `alias` has lexical scope, so they could wrongly believe that maybe "From now on" has a wider scope.

In addition to that, if `Math.List` was implemented within `Math`, then the aliasing is automatic. Thus, the description kind of takes for granted `Math.List` is implemented elsewhere. But since the scope does not match, why should nesting match (the reader may wonder).

The proposal addresses these two points with two minimal changes:

* Adds a comment in the example that indicates the scope of the aliasing, and explicitly says this alias is effective "within the module definition" (from that point on).
* Renames the main module to `Stats`, which is a lateral kick to the second issue.

### DIFF
--- a/getting-started/alias-require-and-import.markdown
+++ b/getting-started/alias-require-and-import.markdown
@@ -1,6 +1,6 @@
 ---
 layout: getting-started
-title: alias, require and import
+title: alias, require, and import
 ---
 
 # {{ page.title }}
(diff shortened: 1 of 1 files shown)


getting-started/alias-require-and-import.markdown
```diff
@@ -82,7 +79,7 @@ Macros are chunks of code that are executed and expanded at compilation time. Th
 
 ```iex
 iex> Integer.is_odd(3)
-** (CompileError) iex:1: you must require Integer before invoking the macro Integer.is_odd/1
+(UndefinedFunctionError) function Integer.is_odd/1 is undefined or private.
```
josevalim: We are missing the `**` prefix here. Btw, in the soon to be out v1.4 the error message is: "** (UndefinedFunctionError) function Integer.is_odd/1 is undefined or private. However there is a macro with the same name and arity. Be sure to require Integer if you intend to invoke this macro".

Shall we use that?

fxn: Awesome, out now, will update the patch later!

fxn: @josevalim updated. I have seen the line gets wrapped and reads well.

josevalim:

PR #901, opened 2016-12-26
Replace deprecated :crypto.md5 call with :crypto.hash(:md5, str)

cabaret:
I kept getting this warning in iex:

`warning: crypto:md5/1 is deprecated and will be removed in a future release; use crypto:hash/2`

The documentation on the front page uses `crypto.md5`. I changed this in favour of `crypto.hash`.


josevalim:

PR #907, opened 2017-01-01
Edits in sigils.md

fxn:
Edits that need some explanation:

* The reason for alternative delimiters is explained with a slightly different emphasis. Point is to be able to write literals that read literally, to avoid escaping delimiters. For example, contrary to what the text says, parentheses for regular expressions are fine as long as the literal has no parentheses.

* In "Strings, char lists and words sigils", "words" becomes "word lists" to match the subsection title. In particular, in the TOC they read next to each other.

* When introducing `~s`, there is no point in mentioning single quotes. You want to change the delimiter if double quotes are present. Presence of single quotes is irrelevant.

* Regarding escape codes, I think `\"` and `\n` belong to different categories. You may leverage that part to talk about escaping delimiters though. Indeed, in you change the delimiter to be parentheses, then `\(` is a parenthesis. That is a escaped delimiter, different than encoding a tab character.

### DIFF
--- a/getting-started/sigils.markdown
+++ b/getting-started/sigils.markdown
@@ -9,7 +9,7 @@ title: Sigils
 
 We have already learned that Elixir provides double-quoted strings and single-quoted char lists. However, this only covers the surface of structures that have textual representation in the language. Atoms, for example, are mostly created via the `:atom` representation.
 
-One of Elixir's goals is extensibility: developers should be able to extend the language to fit any particular domain. Computer science has become such a wide field that it is impossible for a language to tackle many fields as part of its core. Rather, our best bet is to make the language extensible, so developers, companies and communities can extend the language to their relevant domains.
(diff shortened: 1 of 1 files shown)


getting-started/sigils.markdown
```diff
@@ -51,15 +51,15 @@ So far, all examples have used `/` to delimit a regular expression. However sigi
 ~r<hello>
 ```
 
-The reason behind supporting different delimiters is that different delimiters can be more suited for different sigils. For example, using parentheses for regular expressions may be a confusing choice as they can get mixed with the parentheses inside the regex. However, parentheses can be handy for other sigils, as we will see in the next section.
+The reason behind supporting different delimiters is to provide a way to write literals without escaped delimiters. For example, a regular expression with forward slashes like `~r(^https?://)` reads arguably better than `~r/^https?:\/\/`. Similarly, if the regular expression has forward slashes and capturing groups, you may then choose double quotes instead of parentheses.
```
whatyouhide: I would say "... if the regular expression has forward slashes and capturing groups (that use `()`), you may choose ..." just to clarify what `()` have to do with capturing groups. Wdyt?

fxn:
@whatyouhide good suggestion, I have added it.

josevalim:
@whatyouhide please go ahead and merge when you are ready :)

PR #909, opened 2017-01-02
Final minor details

fxn:
I have completed the getting started guide, here's a last patch with minor details.

Things that deserve a comment:

* When I read the paragraph that says that the digraph API modifies the graph in-place I wondered what is this vodoo :). Given this API is like nothing you've seen before, I believe a short explanation would leave the reader less puzzled. I have reordered the exposition to come after the shell session, and leveraged that ETS comes next to explain and connect in one shot.

* "ZIP" is the most common spelling I'd say. For example look at its [MIME registration](http://www.iana.org/assignments/media-types/application/zip). [Wikipedia uses ".ZIP"](https://en.wikipedia.org/wiki/Zip_(file_format)), with a leading dot, but in any case you do not write "zip files" normally. Different from zlib, whose [own documentation](http://www.zlib.net/) consistently writes it in lowercase.

* In the typespecs guide the patch reorders the list to match the order of the following sections.

### DIFF
--- a/getting-started/erlang-libraries.markdown
+++ b/getting-started/erlang-libraries.markdown
@@ -29,7 +29,7 @@ iex> :binary.bin_to_list "Ø"
 [195, 152]
 ```
 
-The above example shows the difference; the `String` module returns UTF-8
+The above example shows the difference; the `String` module returns Unicode
 codepoints, while `:binary` deals with raw data bytes.
 
 ## Formatted text output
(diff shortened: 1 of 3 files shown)


getting-started/erlang-libraries.markdown
```diff
@@ -29,7 +29,7 @@ iex> :binary.bin_to_list "Ø"
 [195, 152]
 ```
 
-The above example shows the difference; the `String` module returns UTF-8
+The above example shows the difference; the `String` module returns Unicode
```
josevalim: :+1:

josevalim:

PR #914, opened 2017-01-05
update current year of copyright to 2017

tony612:

PR #915, opened 2017-01-05
Dynamically update year based on last site build

eksperimental:
So we don't need to update the footer every new year.
Improved version of https://github.com/elixir-lang/elixir-lang.github.com/pull/914

Additionally replaced the hyphen with an en-dash to separate the years
(that's the proper way of writing years lapses).

/cc @tony612


josevalim:

PR #918, opened 2017-01-07
Update mix.exs of guide to use :extra_applications

harfangk:
Mix 1.4 introduced automatic inference of required applications from
dependencies list. Now Mix creates `:extra_applications` instead of
`:applications` by default in application configuration in `mix.exs`.
This commit updates the example `mix.exs` configurations in the guide to
reflect that change.

Signed-off-by: Bonghyun Kim <bonghyun.d.kim@gmail.com>

### DIFF
--- a/getting-started/mix-otp/dependencies-and-umbrella-apps.markdown
+++ b/getting-started/mix-otp/dependencies-and-umbrella-apps.markdown
@@ -169,7 +169,7 @@ defmodule KVServer.Mixfile do
   end
 
   def application do
-    [applications: [:logger],
+    [extra_applications: [:logger],
      mod: {KVServer, []}]
   end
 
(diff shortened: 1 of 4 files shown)


harfangk:
There are two other potential cases.

https://hexdocs.pm/mix/1.4.0/Mix.Tasks.Compile.App.html#content
`def application do
  [mod: {MyApp, []},
   env: [default: :value],
   applications: [:crypto]]
end
`
http://elixir-lang.org/getting-started/erlang-libraries.html#the-crypto-module
> `def application do
  [applications: [:crypto]]
end`

I didn't include them in the PR because they are not incorrect, but would like to bring them to attention.

josevalim:
Thank you, I will look into the other two!

josevalim:

PR #928, opened 2017-01-22
Add "group" key to github-pages gem in Gemfile

eksperimental:
As suggested in Official Github-pages gem article
https://help.github.com/articles/setting-up-your-github-pages-site-locally-with-jekyll/#step-3-optional-generate-jekyll-site-files


josevalim:
It is fine as is since we don't plan to use this repository for anything else. :)

PR #929, opened 2017-01-22
Config repository key

eksperimental:
Add "repository" field in _config.yml (makes use of https://github.com/jekyll/github-metadata)
and updates gems

PR #930, opened 2017-01-22
Keep latest Elixir versions in a data file

eksperimental:
We keep all the latest versions in one file, so we don't need to update them many places.
Additionally it adds links to the Elixir v1.0 docs

### DIFF
--- a/docs.markdown
+++ b/docs.markdown
@@ -61,3 +61,12 @@ Choose which version you want documentation for.
 * [IEx](https://hexdocs.pm/iex/{{ site.data.elixir_version.v1_1.version }}/) - interactive shell
 * [Logger](https://hexdocs.pm/logger/{{ site.data.elixir_version.v1_1.version }}/) - built-in Logger
 * [Mix](https://hexdocs.pm/mix/{{ site.data.elixir_version.v1_1.version }}/) - build tool
+
+<h4 id="v1.0">v1.0</h4>
+
+* [Elixir](https://hexdocs.pm/elixir/{{ site.data.elixir_version.v1_0.version }}/) - standard library
+* [EEx](https://hexdocs.pm/eex/{{ site.data.elixir_version.v1_0.version }}/) - templating library
+* [ExUnit](https://hexdocs.pm/ex_unit/{{ site.data.elixir_version.v1_0.version }}/) - unit test library
+* [IEx](https://hexdocs.pm/iex/{{ site.data.elixir_version.v1_0.version }}/) - interactive shell
+* [Logger](https://hexdocs.pm/logger/{{ site.data.elixir_version.v1_0.version }}/) - built-in Logger
+* [Mix](https://hexdocs.pm/mix/{{ site.data.elixir_version.v1_0.version }}/) - build tool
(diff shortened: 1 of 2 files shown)


docs.markdown
```diff
@@ -26,38 +26,47 @@ Choose which version you want documentation for.
...
+* [Mix](https://hexdocs.pm/mix/{{ site.data.elixir_version.v1_1.version }}/) - build tool
+
+<h4 id="v1.0">v1.0</h4>
+
+* [Elixir](https://hexdocs.pm/elixir/{{ site.data.elixir_version.v1_0.version }}/) - standard library
+* [EEx](https://hexdocs.pm/eex/{{ site.data.elixir_version.v1_0.version }}/) - templating library
+* [ExUnit](https://hexdocs.pm/ex_unit/{{ site.data.elixir_version.v1_0.version }}/) - unit test library
+* [IEx](https://hexdocs.pm/iex/{{ site.data.elixir_version.v1_0.version }}/) - interactive shell
+* [Logger](https://hexdocs.pm/logger/{{ site.data.elixir_version.v1_0.version }}/) - built-in Logger
+* [Mix](https://hexdocs.pm/mix/{{ site.data.elixir_version.v1_0.version }}/) - build tool
```
whatyouhide: Should we loop over all versions then instead of manually listing them?

eksperimental: well, I'm afraid future versions will add new components.
for example 1.0 has no docs

eksperimental:
Forgot to mention that in order to work locally, it needs these changes https://github.com/elixir-lang/elixir-lang.github.com/pull/929

docs.markdown
whatyouhide: Ok but so correct me if I'm wrong: this way we're gonna have to update both `elixir_version.yml` as well as this file, right? I am failing to see the benefit of this, sorry :(

josevalim:
@eksperimental what is missing to make all of them consistent? If it is just the fact that v1.0 does not have a Docs.zip, let's remove it from the listing. If we cannot fully automatize this, then I believe @whatyouhide is correct that it is not really worth it. :) Thank you!

eksperimental:
I don't think we should remove links to docs because the release didn't have a Doc.zip file. I was going to suggest to update the release if that's possible and include it for consistency.
But removing it from the official docs is like pretended 1.0 never existed. There are books written for this version, and if they don't appear in the official documentation page, I don't know where else should we expect them to appear.

regarding what's the point of this PR if we are to duplicate efforts.
1. on a patch release, it's just a matter of updating one file (elixir_version.yml),
2. on a major release it is when you need to manually include the new set of documents in the docs page

let me know what you want to do.

eksperimental:
I will work on a solution that will contemplate all what we need

josevalim:
So let's keep things as is then. Once a new major/minor is out, it is quite unlikely to release a previous version. It has never happened as far as I know. So I will prefer to keep the system simple for updates. Thank you.

eksperimental:
thank you for your input @whatyouhide and @josevalim 
it has all been addressed in  #931

PR #937, opened 2017-01-29
Correct `KV.Supervisor` call to `init`

bspatafora:
- Change `KV.Supervisor` call from `start_link` to `init`
- Correct an "a" to an "an"

### DIFF
--- a/getting-started/mix-otp/supervisor-and-application.markdown
+++ b/getting-started/mix-otp/supervisor-and-application.markdown
@@ -116,7 +116,7 @@ We can also configure the generated `.app` file by customizing the values return
 
 ### Starting applications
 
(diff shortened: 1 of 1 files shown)


getting-started/mix-otp/supervisor-and-application.markdown
```diff
@@ -189,7 +189,7 @@ defmodule KV do
   use Application
 
   def start(_type, _args) do
-    KV.Supervisor.start_link
```
josevalim: `start_link` is correct. `init` is the callback that is invoked by `start_link`, inside the supervisor process.

bspatafora: Whoops, had miscopied the function name as `init` in my `KV.Supervisor`. Sorry!

bspatafora:
My mistake

PR #942, opened 2017-02-02
Clarify keywords-and-maps.markdown

### LINKED ISSUE #941: keywords-and-maps use syntax for referring to first argument of a function
But this is not introduced before.

http://elixir-lang.org/getting-started/keywords-and-maps.html#nested-data-structures

Right after

```
The update_in/2 macro is similar but allows us to pass a function that controls how the value changes. For example, let’s remove “Clojure” from Mary’s list of languages:
```

nhooyr:
Closes #941


josevalim:

PR #946, opened 2017-02-04
fix typo in example.

rkazak:
minor typo in example.


eksperimental:
Your example is not valid Elixir code. You probably meant `list ++ [{:c, 3}]`
and in the line before it clearly states that it can deal with a keyword list.
> For example, we can use ++ to add new values to a keyword list:
so there is nothing wrong with it.

rkazak:
Got it, I think I misread the example.

eksperimental:
Sure. No worries

PR #948, opened 2017-02-09
Add a note about using a branch for a git dependency

baldmountain:
I guessed this would work and it seemed to.


josevalim:
Thank you @baldmountain but we don't need a full reference, so I believe the current examples are enough. Plus we link to the documentation right after, which covers all supported options. :)

baldmountain:
True, but using a GitHub link is common enough that it might be a good idea to include it. But I'm OK either way. :)

PR #949, opened 2017-02-10
updated Maps section

leothorp:
This doc implies that the pipe syntax for updating a Map (like so: `%{map | :a => 3}`) only works for atom keys, but it actually can be used for any key type.  The PR updates the doc to reflect that.  I also added an example usage of `Map.put` to the section on the Map module's functions, because adding a new key/value seemed like one of the more fundamental things someone would wonder how to do.


josevalim:

PR #950, opened 2017-02-13
rename "Modules" page to "Modules and Functions"

Gazler:
Some users were finding it difficult to know where functions page of the
getting started guide was at a glance. Renaming it to "Modules and
Functions" helps to clarify this.

Only the title of the page has been changed so that the URL remains
unchanged.

### DIFF
--- a/getting-started/modules.markdown
+++ b/getting-started/modules.markdown
@@ -1,6 +1,6 @@
 ---
 layout: getting-started
-title: Modules
+title: Modules and Functions
 ---
 
 # {{ page.title }}
--- a/_data/getting-started.yml
+++ b/_data/getting-started.yml
@@ -22,7 +22,7 @@
     - title: Keywords and maps
       slug: keywords-and-maps
 
-    - title: Modules
+    - title: Modules and Functions
       slug: modules
 
     - title: Recursion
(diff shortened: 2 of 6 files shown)


getting-started/modules.markdown
```diff
@@ -1,6 +1,6 @@
 ---
 layout: getting-started
-title: Modules
+title: Modules and Functions
```
whatyouhide: Should we rename this file also? And find a way to redirect.

eksperimental: @whatyouhide I think we need to rename the file to `modules-and-functions.markdown` and add this line in the frontmatter `redirect_from: /getting_started/modules.html`

Gazler: @eksperimental Thanks for the tip. @whatyouhide I have added it as a separate commit so you can choose if you want the redirect or not.

eksperimental: @Gazler you also need to update any references to `modules.html` as in 
https://github.com/elixir-lang/elixir-lang.github.com/blob/779d43c077bfa73028d03983175cef0a4a70d214/getting-started/introduction.markdown#L67

Gazler: @eksperimental done.
```
$ grep -ir "modules.html" getting-started/
getting-started/modules-and-functions.markdown:redirect_from: /getting-started/modules.html
```

eksperimental:
@josevalim merged! 
Wow. that's cool. thanks the for addition!
First PR merged ever on GitHub. I hope it went as planned.

PR #953, opened 2017-02-19
updated test case

beenotung:
This change solves the race problem.
It makes sure the reference is made before the process dies.

### DIFF
--- a/getting-started/mix-otp/supervisor-and-application.markdown
+++ b/getting-started/mix-otp/supervisor-and-application.markdown
@@ -240,10 +240,10 @@ test "removes bucket on crash", %{registry: registry} do
   {:ok, bucket} = KV.Registry.lookup(registry, "shopping")
 
   # Stop the bucket with non-normal reason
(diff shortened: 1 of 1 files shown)


whatyouhide:
I realize the current code may be a bit hostile, but it's worth noting that you can most definitely monitor a dead process: you will just get the `:DOWN` message right when you monitor it and you'll get reason `:noproc`.

```elixir
iex> pid = spawn fn -> :ok end
#PID<0.85.0>
iex> Process.monitor(pid) # "pid" is long dead here
#Reference<0.0.2.1188>
iex> flush()
{:DOWN, #Reference<0.0.2.1188>, :process, #PID<0.85.0>, :noproc}
:ok
```

Quoting the Erlang documentation:

>  Monitors are fired when the monitored process or port terminates, does not exist at the moment of creation, or if the connection to it is lost.

josevalim:
Yup, you should have no race but the proposed code is more readable, so thank you!

josevalim:

PR #956, opened 2017-03-14
Added final delimiter to a regular expression

clearjs:


getting-started/sigils.markdown
```diff
@@ -51,7 +51,7 @@ So far, all examples have used `/` to delimit a regular expression. However sigi
 ~r<hello>
 ```
 
-The reason behind supporting different delimiters is to provide a way to write literals without escaped delimiters. For example, a regular expression with forward slashes like `~r(^https?://)` reads arguably better than `~r/^https?:\/\/`. Similarly, if the regular expression has forward slashes and capturing groups (that use `()`), you may then choose double quotes instead of parentheses.
+The reason behind supporting different delimiters is to provide a way to write literals without escaped delimiters. For example, a regular expression with forward slashes like `~r(^https?://)` reads arguably better than `~r/^https?:\/\//`. Similarly, if the regular expression has forward slashes and capturing groups (that use `()`), you may then choose double quotes instead of parentheses.
```
kelvinst: Ooops. :rofl:

eksperimental: I think it will be a good idea to add a new line after every period. That way when we modify a sentence in the future it doesnt affect the git history of the other two

josevalim:

PR #970, opened 2017-04-12
Replace the term 'dictionary' with 'keyword list'

dlcmh:
I might be mistaken, but the term 'dictionary' seems more appropriate when discussing the relevant Python data structure, while 'keyword list' was returned when performing the following in Iex:

```elixir
{a, b} = {:ok, bucket: "bucket"}
{:ok, [bucket: "bucket"]}

i b
Term
  [bucket: "bucket"]
Data type
  List
Description
  This is what is referred to as a "keyword list". A keyword list is a list
  of two-element tuples where the first element of each tuple is an atom.
Reference modules
  Keyword, List
Implemented protocols
  Collectable, Enumerable, IEx.Info, Inspect, List.Chars, String.Chars
```

### DIFF
--- a/getting-started/mix-otp/agent.markdown
+++ b/getting-started/mix-otp/agent.markdown
@@ -131,7 +131,7 @@ end
 
 We have first defined a setup callback with the help of the `setup/1` macro. The `setup/1` callback runs before every test, in the same process as the test itself.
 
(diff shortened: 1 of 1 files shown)


whatyouhide:
It's definitely a kw list and not a dictionary :) Good catch!

josevalim:

PR #971, opened 2017-04-14
Update elixir version from 1.4.0 to 1.4.1

gopalindians:


josevalim:
Thank you but any Elixir v1.4.0 version should be fine for the tutorial. :)

PR #972, opened 2017-04-14
Update elixir version to 1.4.0

gopalindians:
The tutorial says the minimum requirement of elixir is `1.4.0` and again it says when you enter `iex`  in the terminal, the terminal shows erlang and elixir versions, but in this case(in the docs) it is showing `1.3.4`, which does not met the minimum requirement. So it must at least `1.4.0`. It's good for beginners.


josevalim:

PR #979, opened 2017-04-30
Changes focus in the require explanation

fxn:
### Purpose

This edit syncs the Getting Started guide with https://github.com/elixir-lang/elixir/pull/6041.

### Rationale

To summarize the rationale here: If module `M` wants to use macros from `N`, it needs to require `N` **regardless** of whether `N` is loaded in the VM (see example in the PR linked above). Therefore, ensuring the module is loaded is not really the purpose of `require`, you may have the module loaded and still not be able to access its macros.

That is because `require N` actually adds `N` to a list of modules required by `M`, and being in that list is what controls their visibility. Additionally, the compiler needs to load `N` in order to expand its macros if used, but that is more of an implementation detail.

### Questions

By removing emphasis on the compilation and loading of the required module, the documentation of `use` gets shorter. Indeed, if the reader sees the macro expands to

```elixir
require Foo
Foo.__using__(...)
```

in the current docs they may still wonder what is the point of `require`? With the parallel compiler any function call works at that level and Elixir does whatever necessary to be able to execute it.

So it seems this would be reduced to: `use` enables access to macros and invokes `__using__`. Which made me think if the reader may perceive both things as arbitrary, why couple two apparently unrelated things in one macro?

What do you think?

### DIFF
--- a/getting-started/alias-require-and-import.markdown
+++ b/getting-started/alias-require-and-import.markdown
@@ -13,7 +13,7 @@ In order to facilitate software reuse, Elixir provides three directives (`alias`
 # Alias the module so it can be called as Bar instead of Foo.Bar
 alias Foo.Bar, as: Bar
 
-# Ensure the module is compiled and available (usually for macros)
+# Require the module in order to use its macros
 require Foo
 
 # Import functions from Foo so they can be called without the `Foo.` prefix
@@ -73,9 +73,9 @@ In the example above, since we are invoking `alias` inside the function `plus/2`
 
 ## require
 
-Elixir provides macros as a mechanism for meta-programming (writing code that generates code).
+Elixir provides macros as a mechanism for meta-programming (writing code that generates code). Macros are expanded at compile time.
 
-Macros are chunks of code that are executed and expanded at compilation time. This means, in order to use a macro, we need to guarantee its module and implementation are available during compilation. This is done with the `require` directive:
+Public functions in modules are globally available, but in order to use macros, you need to opt-in by requiring the module they are defined in.
 
 ```iex
 iex> Integer.is_odd(3)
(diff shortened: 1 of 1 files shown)


getting-started/alias-require-and-import.markdown
```diff
@@ -75,7 +75,7 @@ In the example above, since we are invoking `alias` inside the function `plus/2`
 
 Elixir provides macros as a mechanism for meta-programming (writing code that generates code).
 
-Macros are chunks of code that are executed and expanded at compilation time. This means, in order to use a macro, we need to guarantee its module and implementation are available during compilation. This is done with the `require` directive:
+Public functions in modules are globally available, but in order to use macros, you need to opt-in by requiring the module they are defined in.
```
whatyouhide: We're losing the small description about macros here though. Should we keep it? 🤔

fxn: At this point, the guide has not yet covered macros.

I thought that the description in the first paragraph gives the idea to the reader, and it is enough for the purposes of this section. Explaining at which point they are expanded seemed too detailed, the kind of information you cover in the macros chapter proper. Here, in my mind that sentence is there to motivate the rationale being replaced (they are expanded at compilation time, therefore you need to make sure the module is loaded at compilation time, therefore `require`).

That's why the patch removes it, but if you prefer it back I'll add it in a moment.

josevalim: Maybe we can mention it at the end? "by requiring the module they are defined in. Once the module is required, we can invoke macros which are expanded at compilation time." or something of sorts?

fxn: I tried it:

> Public functions in modules are globally available, but in order to use macros, you need to opt-in by requiring the module they are defined in. Once the module is required, we can invoke macros which are expanded at compilation time.

but it kind of says twice that by requiring a module you can access its macros, doesn't it.

What about moving that remark to the first paragraph, which is intended to introduce macros?

> Elixir provides macros as a mechanism for meta-programming (writing code that generates code). Macros are expanded at compile time.

That one is in the current patch.

WDYT?

josevalim: :+1:

josevalim:

PR #983, opened 2017-05-15
Length is not a verb

eksperimental:


getting-started/protocols.markdown
```diff
@@ -9,7 +9,7 @@ title: Protocols
 
 Protocols are a mechanism to achieve polymorphism in Elixir. Dispatching on a protocol is available to any data type as long as it implements the protocol. Let's see an example.
 
-In Elixir, we have two verbs for checking how many items there are in a data structure: `length` and `size`. `length` means the information must be computed. For example, `length(list)` needs to traverse the whole list to calculate its length. On the other hand, `tuple_size(tuple)` and `byte_size(binary)` do not depend on the tuple and binary size as the size information is precomputed in the data structure.
+In Elixir, we have two idioms for checking how many items there are in a data structure: `length` and `size`. `length` means the information must be computed. For example, `length(list)` needs to traverse the whole list to calculate its length. On the other hand, `tuple_size(tuple)` and `byte_size(binary)` do not depend on the tuple and binary size as the size information is precomputed in the data structure.
```
josevalim: `functions` maybe?

eksperimental: Well, no. Because "functions"  refer to a function named `size` and `length` and what we are trying to convey is that these words are used for different situations, and for examples `tuple_size` is not a function named `size` but a function that uses the idiom `size`.
@lpedromachado  [suggested](https://github.com/elixir-lang/elixir-lang.github.com/pull/985) to use "nouns" which is fine to me. I would go with "nouns" or "idioms", either one is fine to me.

whatyouhide: On second thought, @eksperimental's comment makes sense to me. I think "idioms" is indeed the right word because it expresses that "length" and "size" are concepts, not strictly functions. Wdyt @josevalim?

josevalim:

PR #984, opened 2017-05-15
Minor fix in processes.markdown

lpedromachado:
Put "inspect" between "`" and "`".

### DIFF
--- a/getting-started/processes.markdown
+++ b/getting-started/processes.markdown
@@ -90,7 +90,7 @@ iex> receive do
 "Got hello from #PID<0.48.0>"
 ```
 
-The inspect function is used to convert a data structure's internal representation into a string, typically for printing. Notice that when the `receive` block gets executed the sender process we have spawned may already be dead, as its only instruction was to send a message.
+The `inspect` function is used to convert a data structure's internal representation into a string, typically for printing. Notice that when the `receive` block gets executed the sender process we have spawned may already be dead, as its only instruction was to send a message.
 
 While in the shell, you may find the helper `flush/0` quite useful. It flushes and prints all the messages in the mailbox.
 
--- a/getting-started/processes.markdown
+++ b/getting-started/processes.markdown
@@ -90,7 +90,7 @@ iex> receive do
 "Got hello from #PID<0.48.0>"
 ```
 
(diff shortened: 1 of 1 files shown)


getting-started/processes.markdown
```diff
@@ -90,7 +90,7 @@ iex> receive do
 "Got hello from #PID<0.48.0>"
 ```
 
-The inspect function is used to convert a data structure's internal representation into a string, typically for printing. Notice that when the `receive` block gets executed the sender process we have spawned may already be dead, as its only instruction was to send a message.
+The `inspect` function is used to convert a data structure's internal representation into a string, typically for printing. Notice that when the `receive` block gets executed the sender process we have spawned may already be dead, as its only instruction was to send a message.
```
whatyouhide: Can you also add `/0` while you're at it? :) As in `inspect/0`.

eksperimental: Just a question of style. When we refer to to "`receive` block", should we add the arity, as in "`receive/1` block". I was reviewing other Elixir documentation and I came accross this. I went with the arity, but I just wanted to know your opinion, thank you.

whatyouhide: I believe not using the arity makes things like `receive` clearer. They're special forms so I believe it's fine to use `case`, `receive`, `cond`, and friends without the arity.

whatyouhide:
Thanks @lpedromachado! 💟

getting-started/processes.markdown
eksperimental: what's the reason behind special forms being different @whatyouhide

whatyouhide: I just think `case/2` reads worse than `case`. Also, `case` is just particular in that it takes exactly 2 arguments: what would `for` or `with` look like then, since they take a variable number of arguments?

eksperimental: yes, no question about `case` being more legible than `case/2`, but I was just wondering what would be more correct (if that's the word).
the good thing is that special forms have no function with two different arities, so if you refer to `case` it is not ambiguous which `case` it is. On the other hand using arity links to the function documentation. 
But my question in particular was about "`case` blocks" and not when we refer to the special form itself, but as in this case acting as an adjective of "blocks".  So i guess I also prefer "`case` blocks" over "`case/2` blocks". 
Thank you for the discussion @whatyouhide

PR #987, opened 2017-05-30
Fix Secure font loading

OvermindDL1:
Currently the Google fonts link uses `http:`, even when on a secure URL, this causes the font to not be loaded when on a secure URL.  Currently the primary site 'works around' this by having Cloudflare perform URL rewriting, which hides it, however this does not fix mirrors that use direct hosting.  This commit fixes this by allowing the browser to use http or https as necessary as was done with the analytics javascript link earlier, this should complete the work.


josevalim:

PR #988, opened 2017-06-05
Update dependencies-and-umbrella-apps.markdown

ashean:
Minor tweak to the wording. From:
_The caution here is to make sure the domains don't depended on each other (also known as cyclic dependencies)._
To:
_The caution here is to make sure the domains don't each depend on the other (also known as cyclic dependencies)._


eksperimental:
@ashean "don't each" doesn't sound right to me.
I think the original phrase is pretty clear.

ashean:
The issue I was trying to correct was the the "don't depend**ed** on each other". Depended is past tense so isn't correct in this situation. Actually, just dropping the "ed" yields a correct sentence:
The caution here is to make sure the domains don't depend on each other (also known as cyclic dependencies).

josevalim:
@ashean awesome. can you change the PR to just drop the "ed" then? Thank you!

eksperimental:
@ashean you are right. Yes, please just drop the "ed" then. cheers

ashean:
Thats done now.

eksperimental:
thank you @ashean

PR #990, opened 2017-06-09
Create task-and-gen-tcp.markdown

kimshrier:
The KVServer.accept/1 function needs to have an integer port number and System.get_env/1 returns the environment variable value as a string.  The reason you need an integer port number is because :gen_tcp.listen/2 calls :inet_tcp.getserv/1 and :inet_tcp.getserv/1 is expecting the port to be either an integer or an atom.  In the case that the port is an atom, it is used to look up the port number.  For this example code, since we are already assuming an integer port number, just expecting a string version of the integer port number is sufficient.

Also, by passing the entire expression into String.to_integer/1, if the environment variable is missing, you get the missing $PORT message.  If the environment variable is present but not a string representation of an integer, you get an ArgumentError that shows the problem being associated with :erlang.binary_to_integer/1, which is a clear indication of the problem.


josevalim:

PR #991, opened 2017-06-13
Grammer fix in "Static code analysis" section

damonvjanis:
An alternative way to fix the grammar would be to change the word "wrote" to "write", instead of changing the word "if" to "though".


josevalim:

PR #992, opened 2017-06-14
add broken image on learning page

kirkins:

PR #993, opened 2017-06-14
remove image and point to external host

kirkins:
#992


eksperimental:
thank you @krikins
Just to be clear, who owns this server? The thing is previously the link to the course was changed, and now the image, we need to place it in a place where we know will stay permanently. thank you again

kirkins:
@eksperimental the server is the same one linked to in the documentation. The original broken link was to someones personal dropbox.

kirkins:
See line 88 where the book is linked to -> https://goo.gl/9w33lp

josevalim:
Let's wait for a reply from @robconery please.

josevalim:
Rob is ok with it, so let's host it ourselves as it is more fail proof.

robconery:
Thank you! I can put the image online somewhere else if need be...

PR #994, opened 2017-06-14
Add instruction to run command with additional arg

arecvlohe:
On line `221` it is proposed how to start the application with a default port. However, on line `245`when the application is run again, it does not include how to configure the port when calling `mix run --no-halt`.  I suggest updating the command to `PORT=4040 mix run --no-halt`, if that is an adequate solution.


josevalim:

PR #995, opened 2017-06-19
Add new learning resource - elixir flashcards \0/

tetiana12345678:


josevalim:
Awesome! Btw, we have been mostly keeping the list ordered according to when they are added, to avoid debates, so could you please make sure the cards are the last item in the section they belong to? Thank you!

tetiana12345678:
oops. I got it wrong. I grouped flashcards together with books. Sorting function is not by paper 😆 Let me update

tetiana12345678:
Do I need to squash commits?

josevalim:

PR #996, opened 2017-06-23
Add Elixir Koans as a learning resource.

Maikon:
PR to add the [Elixir Koans](http://elixirkoans.io) as a learning resource to the website.


josevalim:

eksperimental:
The image is 201 KB, and it is 1881 px wide when we are only displaying 190px. It seems a lot for such a simple image (four or five colours).
Additionally you can use pngcrush or any online service such as https://tinypng.com/
to compress it even more.
That will improve loading times considerably

josevalim:
Thanks @eksperimental. I force pushed this particular change and compressed the image. Then I went ahead and compressed all images on learning, we got to save about 300kb!

eksperimental:
That's great, I see the image is now 17kb

PR #999, opened 2017-07-27
Fix asciinema spelling

sickill:
I just saw Elixir 1.5 announcement, and the use of asciinema in the post. Awesome!
I noticed that you spelled it "asciicinema" instead of "asciinema", so here's a quick fix.


josevalim:

PR #1003, opened 2017-07-29
Remove inconsistent reference to fault tolerance from Tasks and gen_tcp chapter (#1002)

### LINKED ISSUE #1002: Inconsistency within `Task and gen_tcp` section
Hello,

I'm raising this as an issue since I don't know how you'd prefer to resolve this, but the recent 1.5 update resulted in an inconsistency within the "Task and gen_tcp" chapter of the the Mix and OTP guide, in particular these two sections:

```elixir
  def start(_type, _args) do
    children = [
      {Task, fn -> KVServer.accept(4040) end}
    ]

    opts = [strategy: :one_for_one, name: KVServer.Supervisor]
    Supervisor.start_link(children, opts)
  end
```

> Yes, it works! If you kill the client, the whole server will still crash. **But you will see that another one starts right away.** However, does it scale?

However, unlike `Supervisor.Spec.worker/3`, because `Task.child_spec/1` defaults to a restart option of `:temporary` and not `:permanent` this restart does not happen.

Like I said, not sure how you'd prefer to change this but the official Task docs suggest defining Tasks within a module and that together with `use Task, restart: :permanent` could be one option.

coryc5:
Closes #1002 

also changes a reference from `init/1` to `start/2`

### DIFF
--- a/getting-started/mix-otp/task-and-gen-tcp.markdown
+++ b/getting-started/mix-otp/task-and-gen-tcp.markdown
@@ -160,7 +160,7 @@ say me
 say me
 ```
 
-Yes, it works! If you kill the client, the whole server will still crash. But you will see that another one starts right away. However, does it *scale*?
+Yes, it works! However, does it *scale*?
 
 Try to connect two telnet clients at the same time. When you do so, you will notice that the second client doesn't echo:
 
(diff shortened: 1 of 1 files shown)


josevalim:

PR #1012, opened 2017-08-14
Invalid string (byte sequence) is no longer false.

dbrown428:
I'm learning Elixir at the moment, so I apologize if I've overlooked something, but this is what I've found:

Elixir 1.5.0
iex> String.valid?(<<239, 191, 191>>
true

Elixir 1.4.0
iex> String.valid?(<<239, 191, 191>>
false

I'm not sure if the documentation should be updated to reflect the new behaviour of String.valid? or if the function is not working the way it was intended to. At this point, I've assumed the documentation needs to be updated to the new behaviour so I just tweaked the byte sequence to be an invalid string.

Any thoughts?


josevalim:
You are correct, the previous String.valid? was marking valid strings as invalid. It has been fixed. :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #1016, opened 2017-08-18
Link to Mix.env/0 instead of Mix.env/1

sonicdoe:
Since the code example uses [`Mix.env/0`](https://hexdocs.pm/mix/Mix.html#env/0) (instead of [`Mix.env/1`](https://hexdocs.pm/mix/Mix.html#env/1)), I think it makes more sense to link to `Mix.env/0`.


josevalim:

PR #1017, opened 2017-08-18
Bug at Documentation: Wrong Error mentioned

chetkhatri:
What were proposed in this Pull Request ?
* iex> length([1, [2], 3]) = 3
above syntax won't give below error 
** (CompileError) iex:1: illegal pattern
but it gives ** (CompileError) iex:1: cannot invoke remote function :erlang.length/1 inside match

so it has been manually verified at shell and update the documentation

### DIFF
--- a/crash-course.markdown
+++ b/crash-course.markdown
@@ -634,13 +634,18 @@ Anonymous functions are first-class values, so they can be passed as arguments t
 **Erlang**
 
 ```erlang
+% math.erl
 -module(math).
 -export([square/1]).
 
 square(X) -> X * X.
+```
 
-lists:map(fun math:square/1, [1, 2, 3]).
-%=> [1, 4, 9]
+```erl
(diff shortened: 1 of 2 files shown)


crash-course.markdown
```diff
@@ -634,13 +634,19 @@ Anonymous functions are first-class values, so they can be passed as arguments t
...
+```
 
-lists:map(fun math:square/1, [1, 2, 3]).
-%=> [1, 4, 9]
+```erl
+Eshell V5.9  (abort with ^G)
+1> c(math).
+{ok,math}
+2> lists:map(fun math:square/1, [1, 2, 3]).
+[1,4,9]
```
whatyouhide: Is this change needed/relevant for this PR?

chetkhatri: Looks, predecessor PR got merged ! let me check

PR #1018, opened 2017-08-18
Wrong Error exception issue at documentation fixed

chetkhatri:
What were proposed in this Pull Request ?

    iex> length([1, [2], 3]) = 3
    above syntax won't give below error
    ** (CompileError) iex:1: illegal pattern
    but it gives ** (CompileError) iex:1: cannot invoke remote function :erlang.length/1 inside match

so it has been manually verified at shell and update the documentation.


josevalim:

PR #1024, opened 2017-08-27
Language edit

msimonborg:
"It is the time we change that" is an awkward phrase, subbed it out for "It is time we change that"


josevalim:

PR #1027, opened 2017-09-01
Remove stray underscore.

Doerge:


josevalim:
@Doerge it is by choice to signal that the variable is not used - otherwise a warning is emitted. :)

Doerge:
Ah, I see. Wouldn't `_` make more sense then? Or is this a convention in Elixir to explain what goes into the unused var? I'm reading the Getting Started from front to back and have not met this before, so it confused me.

josevalim:
I see. I will check the guides and make sure if we can/should introduce this concept earlier. But yes, you could simply use `_`. But we prefer to still give it a name for clarity.

PR #1029, opened 2017-09-02
[install.markdown] Update Ubuntu and Debian versions

abesto:
* Ubuntu Precise (12.04) has reached EOL, removed from list.
 * Added other Ubuntu and Debian versions with packages available on https://packages.erlang-solutions.com/erlang/


josevalim:

PR #1030, opened 2017-09-02
Re-word discussion of list prepend / append performance

abesto:
* The text mentioned prepending, while the code was concatenating two lists. The distinction between whether "prepending" or "appending" is fast depends on the size of the left-hand list.
 * There exists explicit prepending to a list as well, which is constant time

This is pretty much migrating some content from https://hexdocs.pm/elixir/List.html.

### DIFF
--- a/getting-started/basic-types.markdown
+++ b/getting-started/basic-types.markdown
@@ -349,16 +349,12 @@ What is the difference between lists and tuples?
 
 Lists are stored in memory as linked lists, meaning that each element in a list holds its value and points to the following element until the end of the list is reached. This means accessing the length of a list is a linear operation: we need to traverse the whole list in order to figure out its size.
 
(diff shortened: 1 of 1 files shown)


josevalim:
I agree that the definition of prepending and appending here is confusing and the way you wrote it is very clear: "it depends on on the size of the left side".

However I would like to avoid talking about cons cells this early on. Especially given that `[0 | list]` is semantically equivalent to `[0] ++ list`. As `[0, 1, 2, 3] ++ list` is equivalent to `[0, 1, 2, 3 | list]`. 

What do you think? Thank you for the PR.

abesto:
Yep, I was debating with myself too whether including `|` is worth it. Up to you, I'm fine either way.

josevalim:
Thank you! I have pushed some changes to your branch based on the discussion.

josevalim:

PR #1037, opened 2017-09-19
Update Pathogrean Triples example (missing triples)

uDude:
The last Pythagorean triples is missing numerous Pythagorean triples.  The filter "a + b + c <= n" looks like an accident in the application the triangle inequality.  The theorem actually states "a + b >= c."  Adjusting this will result in the inclusion of missing triples without duplication.


josevalim:

PR #1043, opened 2017-10-02
Update typespecs-and-behaviours.markdown to remove compilation errors in the Parser definition

rheng:
The original documentation was throwing an error because reason and value are not defined. Assuming reason is a string and the value is a term, the proposed changes at least compiles. This also matches the definition in line 145. I also added a note explaining what `term` represents.

Error:
```
** (CompileError) iex:2: type reason() undefined
    (stdlib) lists.erl:1338: :lists.foreach/2
    (stdlib) erl_eval.erl:670: :erl_eval.do_apply/6 
    (iex) lib/iex/evaluator.ex:219: IEx.Evaluator.handle_eval/5
```

### DIFF
--- a/getting-started/typespecs-and-behaviours.markdown
+++ b/getting-started/typespecs-and-behaviours.markdown
@@ -105,12 +105,12 @@ We can create a `Parser` behaviour:
 
 ```elixir
 defmodule Parser do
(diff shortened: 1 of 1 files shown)


josevalim:

PR #1044, opened 2017-10-03
Change KV.Bucket to KV.Registry

dcadenas:
Looks like this is what is meant.


josevalim:
Indeed!  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #1046, opened 2017-10-08
Update task-and-gen-tcp.markdown to have KVServer module

rheng:
Added a KVServer module definition to the file. Received this error without it:
`** (ArgumentError) cannot invoke def/2 outside module`


josevalim:

PR #1054, opened 2017-10-28
Modify wording on the usage of some operators in Lists to make it sound more accurate

boddhisattva:
This PR is an attempt to be more explicit(via a wording fix) on the purpose each operator serves when used in the context of Lists in Elixir


whatyouhide:
Thanks @boddhisattva! :heart_decoration:

boddhisattva:
Thank you @whatyouhide :)

PR #1055, opened 2017-11-01
Fix missing link to QuickCheck paper

thetamind:
* Do we want to badge link as directly linking to a PDF (is this still good etiquette)?
* This looks like the correct paper from ICFP 2000, as the original link [from here](http://www.cse.chalmers.se/~rjmh/QuickCheck/) is broken.


whatyouhide:
Great catch @thetamind, thanks! 💟

PR #1064, opened 2017-12-19
Add missing 'and'

half-cambodian-hacker-man:
"Aliases start in upper case *and* are also atoms"


josevalim:

Issue #1065, opened 2017-12-24
Proposal - Can we consider adding a Debugging Section to the Getting Started guides?


boddhisattva:
As a beginner to Elixir, I started learning the concepts of the language by reading up the Getting Started Guides and also by solving some [exercises in the Elixir track of Exercism](http://exercism.io/languages/elixir/exercises). As I was solving more problems on Exercism, they were times when I wasn't sure how to use `IEx.pry` wrt an Elixir script until I came across [this discussion](https://groups.google.com/forum/#!msg/elixir-lang-talk/8VejRgYw2hQ/eZPNcdMlEQAJ) and then realised, I could now go ahead and add `IEx.pry` to my code and call any part of the same script through Elixir's interactive shell. 

I believe in general many beginners to Elixir could benefit from a `Debugging in Elixir` guide and I'd be happy to work on writing one if we're for the idea of having it as part of the Getting Started with Elixir section. I've also seen a [discussion here](https://groups.google.com/forum/#!topic/elixir-lang-talk/9e7JjfX70NA) on elixir-lang talk mailing list where they talk about a missing Debugging section as part of the Getting Started series.

If there's an interest on having one, I was thinking of covering a few examples related to the usage of:
1. IO.puts
2. IO.inspect
3. IEx.pry

Please feel free to share your thoughts on the above and also please feel free to suggest what you'd like to see covered additionally in a basic guide of Debugging with Elixir for beginners so that I can work accordingly on the documentation for this if we're planning to have one.

Thank you :)

josevalim:
I believe that's a great idea. I don't think we need to include IO.puts, if we have IO.inspect. I would also mention :observer and :debugger coming from Erlang, as well as the new IEx.break!. It can be a new chapter after the typespec and behaviour one.

boddhisattva:
Thank you @josevalim 👍 . I aim to push a WIP PR trying this out for `IO.inspect` and `IEx.pry` for starters just to get some initial feedback based on how I'm planning to structure the documentation wrt this section and proceed with working on mentioning about `:observer` , `:debugger` and `IEx.break!` on an incremental basis based on the received feedback wrt `IO.inspect` and `IEx.pry` related examples. Hope that works, Thank you.

josevalim:
Sounds perfect! Let’s get started with those two and then increment as
necessary.
-- 


*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Director of R&D*

[...]

PR #1067, opened 2017-12-25
Add a debugging chapter as part of the Getting Started with Elixir series

boddhisattva:
Addresses [this issue](https://github.com/elixir-lang/elixir-lang.github.com/issues/1065). Kindly note this PR is work in progress and is submitted in it's current state to get some initial feedback before more work can be done as part of the Debugging chapter.

### DIFF
--- a/_data/getting-started.yml
+++ b/_data/getting-started.yml
@@ -61,6 +61,9 @@
     - title: Typespecs and behaviours
       slug: typespecs-and-behaviours
 
+    - title: Debugging
+      slug: debugging
+
     - title: Erlang libraries
       slug: erlang-libraries
 
--- a/getting-started/debugging.markdown
+++ b/getting-started/debugging.markdown
@@ -1,0 +1,55 @@
+---
+layout: getting-started
+title: Debugging
+---
+
+# {{ page.title }}
+
+{% include toc.html %}
+
+There are a number of ways one can debug their code in Elixir. In this chapter we will cover some of the more common ways of doing so.
+
+## IO.inspect/2
+
+What makes `IO.inspect(item, opts \\ [])` really useful in debugging is that it returns the `item` argument passed to it without affecting the behavior of the original code. Let's see how that is so with an example.
+
+```elixir
+(1..10)
+|> IO.inspect
+|> Enum.map(fn x -> x * 2 end)
+|> IO.inspect
+|> Enum.sum
+|> IO.inspect
+```
+
+Prints:
+```elixir
+1..10
+[2, 4, 6, 8, 10, 12, 14, 16, 18, 20]
+110
+```
+
+As you can see here `IO.inspect/2` makes it possible to "spy" on values almost anywhere in your code like in the above case without altering the result. One of the places where IO.inspect is very helpful is when we have to debug code one or more times inside of a pipeline like in the above case.
+
+Another handy thing that the `IO.inspect/2` function provides is the ability to decorate the output with a `label` option. The label will be printed before the inspected `item`. A sample usecase where this could be handy is shown below.
+
+```elixir
+[1, 2, 3]
+|> IO.inspect(label: "before")
+|> Enum.map(&(&1 * 2))
+|> IO.inspect(label: "after")
+|> Enum.sum
+```
+
+Prints:
+
+```elixir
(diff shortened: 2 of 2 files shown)


boddhisattva:
@josevalim and other collaborators of this project, I've started work on the debugging chapter with introducing `IO.inspect` for starters. Would be great to get your feedback on the contents for wrt `IO.inspect` before I proceed with working on other parts of the debugging chapter. Also, would it be feasible to do incremental merges to this chapter with each new debugger option added(as mentioned [here](https://github.com/elixir-lang/elixir-lang.github.com/issues/1065#issuecomment-353778003)) be a potential independent candidate(the candidates being: `:observer`, `:debugger` & `IEx.break!`)) that could be merged to this repo? This way what content is already ready can be made available to those who would be going through the Elixir getting started guides and they can hopefully find this info useful without having to wait for the whole chapter to be released to benefit from the content. I will most likely continue to work on these sections(one for  :observer, :debugger & IEx.break!) as and when I get time in the coming weeks.

There could be other things I could potentially be missing with incremental merges as well so happy to hear your thoughts to take things forward accordingly :) Thank you :)

josevalim:
@boddhisattva we are a bit busy right now since we were getting 1.6 out of the door but my plan is to review your PR and the guides once things calm down a bit. Thanks!

boddhisattva:
Totally understandable @josevalim 👍 . Thank you for your prompt response :) Happy to wait and follow up again after a few weeks.

juhalehtonen:
I just read through the current WIP covering IO.inspect, and I think @boddhisattva has done a good job with introducing it through two easy-to-understand examples, and at I would be happy to see this made available on the website :)

As for your comment/question regarding incremental merges to this chapter, I started wondering whether it would be good to have a few other topics (at least :observer) covered prior publishing it or not. On the one hand, getting information to the people who read the docs swiftly is valuable in its own right, and I fully agree that having at least _some_ documentation available is better than having none. 

On the other hand, I started wondering whether leaving something as important as :observer out initially can backfire for those who read the docs before the expanded docs would be available, and would then assume that the Elixir debugging experience is limited to just using IO.Inspect without realizing that there's far more to it that just isn't covered. I admit that I only have a sample size of three people to go from, but everyone I know has only read the official docs through once before venturing elsewhere.

This isn't to say I'd be opposed to having the IO.Inspect made available as swiftly as possible (and I once again want to commend you on the work you've put in!), but rather just something that popped up in my head while reading through the commit and your comments :) So, any thoughts on this?

boddhisattva:
@juhalehtonen Thank you for the kind words 🙏 and thank you also for taking the time out to review the content and share your feedback on the same as it helps me validate whether the content could help the intended audience or would I have to make some more changes before I proceed further.

I'm completely with you on the idea of providing at least 1-2 lines about the other debugging options( with relevant links) :). I think it would definitely be more helpful than just mentioning about `IO.inspect` alone as part of the initial version of the Debugging chapter.

I'm also happy to wait for other Elixir core team members to get back on this once they get time and we could definitely plan the next steps accordingly.

Thank you again for sharing your feedback :).

josevalim:
This first draft looks great, thank you! :+1: Thank you!

I definitely think we need to cover `IEx.break! + IEx.pry`, `:observer` and `:debugger`. Also note we explore observer on the Mix and OTP guide. But a quick introduction here would help.

@boddhisattva if you want to tackle those other topics, it will definitely be welcome. Otherwise I will work on them right after merging this. :)

boddhisattva:
@josevalim Thank you for the kind words 🙏 , I'm really glad to know it was worth the effort. I'm still a beginner to Elixir and I would end up taking more time on learning and experimenting with each of the other debugger commands. IMHO I think based on my current work, I wouldn't be able to work on them that swiftly and I believe it's in the best interests of the community to have them out earlier if you can work on them right after merging this as you've mentioned above :)

I'm really glad that there would be a Debugging chapter out that could help more people get up and running with Elixir faster! Thank you for your encouragement and support on this José  :)

josevalim:
Merged manually and pushed, it is live! Thank you @boddhisattva for the great start!

boddhisattva:
Thank you very much @josevalim 🙏  Really happy that this is now live!

PR #1070, opened 2018-01-08
Update comprehensions.markdown

luhanghang:


josevalim:
I believe the current code is correct. We already have that `a < b < c`, so we need to prove that the sum of the two smaller sides of the triangle are <= then the bigger one. Or am I missing something? Thank you!

PR #1071, opened 2018-01-10
Fix bug in the "removes bucket on crash" test

sharils:
With BucketSupervisor implemented, the registry should not crash thus lookup should return the same result

### DIFF
--- a/getting-started/mix-otp/dynamic-supervisor.markdown
+++ b/getting-started/mix-otp/dynamic-supervisor.markdown
@@ -29,7 +29,7 @@ test "removes bucket on crash", %{registry: registry} do
 
   # Stop the bucket with non-normal reason
   Agent.stop(bucket, :shutdown)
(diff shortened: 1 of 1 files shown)


getting-started/mix-otp/dynamic-supervisor.markdown
```diff
@@ -29,7 +29,7 @@ test "removes bucket on crash", %{registry: registry} do
 
   # Stop the bucket with non-normal reason
   Agent.stop(bucket, :shutdown)
-  assert KV.Registry.lookup(registry, "shopping") == :error
+  assert KV.Registry.lookup(registry, "shopping") == {:ok, bucket}
```
josevalim: But the bucket should still crash and the registry should remove it from the lookup table. Hence `:error`. You can try it in an actual application here: https://github.com/josevalim/kv_umbrella

PR #1073, opened 2018-01-10
Clarified Introduction Phrasing

1bardesign:
Just a few slight changes to the introductory paragraph. I got confused while reading through and thought "foundation" was referring to an actual organisation like several other open source foundations. I've clarified the intent of that passage, and hopefully this should avoid any similar future confusion.

I discussed this with user `asonge` on the `#elixir-lang` freenode IRC and they recommended I make a PR, so here it is 🙂

### DIFF
--- a/getting-started/introduction.markdown
+++ b/getting-started/introduction.markdown
@@ -10,7 +10,7 @@ redirect_from: /getting_started/1.html
 
 Welcome!
 
-In this tutorial, we are going to teach you about Elixir - the language syntax, how to define modules, how to manipulate the characteristics of common data structures, and more. This chapter will focus on ensuring that Elixir is installed and that you can successfully run Elixir's Interactive Shell, called IEx.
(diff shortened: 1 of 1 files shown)


getting-started/introduction.markdown
```diff
@@ -10,7 +10,7 @@ redirect_from: /getting_started/1.html
 
 Welcome!
 
-In this tutorial, we are going to teach you the Elixir foundation, the language syntax, how to define modules, how to manipulate the characteristics of common data structures, and more. This chapter will focus on ensuring Elixir is installed and that you can successfully run Elixir's Interactive Shell, called IEx.
+In this tutorial, we are going to teach you about Elixir - the language syntax, how to define modules, how to manipulate the characteristics of common data structures, and more. This chapter will focus on ensuring that Elixir is installed and that you can successfully run Elixir's Interactive Shell, called IEx.
```
whatyouhide: Maybe "the fundamentals" of Elixir?

1bardesign: That'd definitely be an option, but I think simpler is better when it comes to something accessibility-sensitive like the intro tutorial. If others would prefer that wording or any other I'm happy to update the PR tomorrow though 👍

eksperimental: "the building blocks of Elixir" could be another option

josevalim:

PR #1079, opened 2018-01-24
Fix old reference to `simple_one_for_one`

nathanl:
The next page actually covers Dynamic supervisors.


josevalim:

PR #1080, opened 2018-01-25
Fix typo in the code sample for DynamicSupervisors

muscar:
The code in `KV.BucketSupervisor.start_bucket` is calling the `Supervisor.start_child` function when it should (probably) be calling the `DynamicSupervisor.start_child` function. I tried the code sample on my machine, and it fails with a function match error with the former call, but it runs ok with the latter.


josevalim:

PR #1081, opened 2018-01-25
Minor change in output

pickatag:
Minor change in how test results are shown (compared to actual output)

Pls note, I'm a complete git newbie so please tell me if I'm doing something I should not...


josevalim:

PR #1083, opened 2018-01-30
Minor change in Agent chapter

mbirman:
Updated error message to include the actual module name.


whatyouhide:
Thanks @mbirman! 💟

PR #1086, opened 2018-02-01
added KV.BucketSupervisor to documentation

### LINKED ISSUE #1085: Dynamic Supervison page incomplete
On this page "The bucket supervisor" section is incomplete.
There is no KV.BucketSupervisor implementation.

sentyaev:
Closes #1085

### DIFF
--- a/getting-started/mix-otp/dynamic-supervisor.markdown
+++ b/getting-started/mix-otp/dynamic-supervisor.markdown
@@ -52,14 +52,28 @@ We are going to solve this issue by defining a new supervisor that will spawn an
 
 ## The bucket supervisor
 
-Let's define a DynamicSupervisor and give it a name of `KV.BucketSupervisor` in `lib/kv/supervisor.ex` as follows:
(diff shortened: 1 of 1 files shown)


josevalim:
We don't need to define one. We can just start the DynamicSupervisor directly in our supervision tree, without defining a separate module for it. I will make it clearer in the guides, thank you!

josevalim:

sentyaev:
As I said I'm very new to Elixir and probably I doing something wrong, but I've tried follow this guide two times and without introducing `KV.BucketSuperviser` (when I just follow this guide) I receive this errors when I try to do `iex -S mix`:
```
** (Mix) Could not start application kv: KV.start(:normal, []) returned an error: an exception was raised:
    ** (ArgumentError) The module DynamicSupervisor was given as a child to a supervisor
but it does not implement child_spec/1.
```

josevalim:
The guide requires Elixir v1.6.1, you are on Elixir v1.6.0. Your solution would work fine though if you don't want to upgrade. :)

PR #1087, opened 2018-02-01
ETS page

sentyaev:
Changed `KV.BucketSupervisor.start_bucket` with `DynamicSupervisor.star_child`.
As I understand for this documentation with Elixir 1.6.1 `KV.BucketSupervisor` was removed, so it's necessary to update `KV.Registry.handle_call` also.


josevalim:

PR #1091, opened 2018-02-19
Call ExUnit.Callback function correctly.

samacs:
I guess the function is now `start_supervised/2` instead of `start_supervised!/2`.

https://hexdocs.pm/ex_unit/ExUnit.Callbacks.html#start_supervised/2

### DIFF
--- a/getting-started/mix-otp/genserver.markdown
+++ b/getting-started/mix-otp/genserver.markdown
@@ -135,7 +135,7 @@ defmodule KV.RegistryTest do
   use ExUnit.Case, async: true
 
   setup do
-    registry = start_supervised!(KV.Registry)
+    registry = start_supervised(KV.Registry)
     %{registry: registry}
   end
 


whatyouhide:
@samacs `start_supervised/2` returns `{:ok, pid}`, so what we want now is still `start_supervised!/2` I believe.

samacs:
Ohh, I see.

I actually was following the tutorial and got an error about that function not being found in `ExUnit.Case`. I ended up using `{:ok, registry} = start_supervised(KV.Registry)` for that matter.

Thanks anyway!

Using:
```
Erlang/OTP 20 [erts-9.1.1] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]

Elixir 1.5.2
```

josevalim:
You need to use Elixir v1.6.1. :)

PR #1093, opened 2018-02-21
Add explanation for functions without body

mahyarap:
Add explanation for functions without body when used as a way to document the arguments.


josevalim:

PR #1095, opened 2018-02-25
Fix message after Stream.cycle call on lists

RonaldKlaus:
Since `Stream.cycle/1` calls `Stream.unfold/2` for a `list`, the message should be `Stream.unfold/2`.


josevalim:

PR #1100, opened 2018-03-16
Adding explanation about the naming conventions

Afoucaul:
I propose to add short explanation about naming conventions in the "Modules and functions" chapter.
I feel introducing these (namely, trailing question mark and leading underscore) is a bit confusing, and telling the reader that it's just naming conventions should reassure them.


josevalim:

PR #1104, opened 2018-03-22
Adding examples to structs presentation

Afoucaul:
I'm proposing to add some explanation and examples about how to mix keys with an explicit default value and keys defaulting to `nil`.
I added two snippets, one showing that one can explicitly make a field default to `nil`, and the other how one can simply use an atom with no value and have it default to `nil`.

This StackOverflow post helped me: https://stackoverflow.com/questions/41222494/correct-struct-definition

### DIFF
--- a/getting-started/structs.markdown
+++ b/getting-started/structs.markdown
@@ -135,7 +135,7 @@ iex> defmodule Product do
 ...>   defstruct name: nil, price: 12
 ...> end
 iex> %Product{}
-%Product{name: nil}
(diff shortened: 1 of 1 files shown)


josevalim:
@Afoucaul yup! That sounds like the way to go!

josevalim:
Ping. :)

mfrasca:
this was superseded by the recent merged pr #1265 .

josevalim:
Beautiful!

PR #1105, opened 2018-03-23
Update genserver.markdown

weaming:
fix test fail

### DIFF
--- a/getting-started/mix-otp/genserver.markdown
+++ b/getting-started/mix-otp/genserver.markdown
@@ -135,7 +135,7 @@ defmodule KV.RegistryTest do
   use ExUnit.Case, async: true
 
   setup do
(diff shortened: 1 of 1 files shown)


getting-started/mix-otp/genserver.markdown
```diff
@@ -135,7 +135,7 @@ defmodule KV.RegistryTest do
   use ExUnit.Case, async: true
 
   setup do
-    registry = start_supervised!(KV.Registry)
+    {:ok, registry} = start_supervised!(KV.Registry)
```
josevalim: Notice the exclamation mark on `start_supervised!`. This means it will return just the registry and not the tuple. :)

josevalim:
Thanks @weaming! I have posted a comment, the current code should be correct. :)

PR #1107, opened 2018-03-26
Update core team information

michalmuskala:
I found the order to be alphabetic with regards to first name, so I put myself in that order as well - it's possible that was just accidental, though.


josevalim:

PR #1111, opened 2018-04-03
Updated the output of `:ets.new/2` in ETS documentation

Afoucaul:
The type of table identifiers changed from integer to reference in OTP20, according to the changelog : http://erlang.org/download/otp_src_20.0.readme
I edited the first snippet where `:ets.new/2` was returning an integer, so as to have it return a reference.
I was pointed toward the document thanks to this answer on StackOverflow: https://stackoverflow.com/a/49633326/7051394

### DIFF
--- a/getting-started/mix-otp/ets.markdown
+++ b/getting-started/mix-otp/ets.markdown
@@ -21,7 +21,7 @@ ETS allows us to store any Elixir term in an in-memory table. Working with ETS t
 
 ```iex
 iex> table = :ets.new(:buckets_registry, [:set, :protected])
-8207
+#Reference<0.1885502827.460455937.234656>
 iex> :ets.insert(table, {"foo", self()})
 true
 iex> :ets.lookup(table, "foo")


whatyouhide:
I would argue that we don't need to show the output of `:ets.new/2` at all and that we would be better off doing something like

```elixir
iex> table = :ets.new(...)
iex> :ets.insert(table, ...)
...
```

Do we have a precedent of doing this in the guides? We definitely do in doctests.

josevalim:
@whatyouhide :+1:.

Afoucaul:
As a learner, I feel more comfortable with a step-by-step demonstration where I'm not surprised by an output that differs from what I'm told I should get.Additionally, the output of :ets.new/2 might be confusing for the learner, since references are an Erlang concept (if I'm not mistaken?).That's why I find it more pedagogical to actually show the expected output, let alone the consistency with the rest of the guide.But really, I'm just nitpicking to polish an already great doc :)-------- Message original --------Objet : Re: [elixir-lang/elixir-lang.github.com] Updated the output of `:ets.new/2` in ETS documentation (#1111)De : Andrea Leopardi À : "elixir-lang/elixir-lang.github.com" Cc : Armand Foucault ,Author I would argue that we don't need to show the output of :ets.new/2 at all and that we would be better off doing something like
iex> table = :ets.new(...)
iex> :ets.insert(table, ...)
...
Do we have a precedent of doing this in the guides? We definitely do in doctests.

—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or mute the thread.


  
  




{"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/elixir-lang/elixir-lang.github.com","title":"elixir-lang/elixir-lang.github.com","subtitle":"GitHub repository","main_image_url":"https://cloud.githubusercontent.com/assets/143418/17495839/a5054eac-5d88-11e6-95fc-7290892c7bb5.png","avatar_image_url":"https://cloud.githubusercontent.com/assets/143418/15842166/7c72db34-2c0b-11e6-9aed-b52498112777.png","action":{"name":"Open in GitHub","url":"https://github.com/elixir-lang/elixir-lang.github.com"}},"updates":{"snippets":[{"icon":"PERSON","message":"@whatyouhide in #1111: I would argue that we don't need to show the output of `:ets.new/2` at all and that we would be better off doing something like\r\n\r\n```elixir\r\niex\u003e table = :ets.new(...)\r\niex\u003e :ets.insert(table, ...)\r\n...\r\n```\r\n\r\nDo we have a precedent of doing this in the guides? We definitely do in doctests."}],"action":{"name":"View Pull Request","url":"https://github.com/elixir-lang/elixir-lang.github.com/pull/1111#issuecomment-378301083"}}}

josevalim:
Great points as well @Afoucaul!

@whatyouhide i am fine either way, your call.

whatyouhide:
> since references are an Erlang concept (if I'm not mistaken?)

It doesn't reference an Erlang concept in particular. Every Erlang data type is an Elixir data type too (and viceversa).

Anyways, let's try it like this and if someone complains that the return value is confusing or that it's different on their machine, we can change it to remove it.

Thanks @Afoucaul! 💟

PR #1112, opened 2018-04-07
Removed filter preventing any correct result

dasim:


josevalim:
Note we don't want the sum of the sides to exceed `n`. That makes it consistent with the optimized generator that comes later. :)

dasim:
But when I ran the code as is, it didn't return any results at all. I'll check again later so I don't waste your time but I tried it couple of times.

josevalim:
Be sure to pass a big number as argument, such as 12. The examples in the guide also work. :)

PR #1113, opened 2018-04-08
Wrong Error exception in Pattern Matching doc

samgh96:
Calling `_` in iex returns a CompileError noting invalid use of `_` instead of an unbound variable error.


josevalim:

PR #1115, opened 2018-04-10
Added how to get whitespace's code point

Afoucaul:
I was searching how to access the code point of an ASCII whitespace, when I came across [this post](https://stackoverflow.com/a/27004332/7051394) on StackOverflow explaining the `?` code point accessor.
I found the info in a comment by @josevalim, saying that "we should probably add such examples to the documentation", so here you go :)

I added a line about the `?\s` syntax, and a short console demonstrating it.

### DIFF
--- a/getting-started/binaries-strings-and-char-lists.markdown
+++ b/getting-started/binaries-strings-and-char-lists.markdown
@@ -50,6 +50,13 @@ iex> ?ł
 322
 ```
 
+Note that the code point of an ASCII whitespace can be accessed by `?\s`:
+
+```iex
+iex> ?\s
+32
+```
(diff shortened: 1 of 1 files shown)


josevalim:
Hi @Afoucaul! We generally avoid such changes because the goal of the getting started guide is not to be a complete reference. Imagine this is the first contact of many with the language, so we need to balance the amount of information they receive. Thank you!

Afoucaul:
@josevalim Ah, I understand. In this case, how about just add `?\s` to the previous REPL session with `?a` and `?ł`? Then, we'd have the info, but it'd be more like a detail.

PR #1120, opened 2018-05-18
Minor add to "Adopting behaviours" examples

igneel64:
The example modules JSONParser and YAMLParser on the "Adopting behaviours" examples would not compile ( "(ArgumentError) cannot invoke def/2 inside function/macro" ) if the parse function does not return anything. By adding a simple empty string we allow a beginner to copy the code and see the expected behaviour. The thing that matters in these example is for the reader to experiment with leaving out a Parser callback and witness the behaviour warning, not having to figure out a compilation error at the time he is learning.


josevalim:

PR #1121, opened 2018-05-24
Formatting

pdawczak:
Hello!

I've re-read the articles for OTP, and noticed the code is still formatted "the old way". I've spent some time to re-format it the way it would be left by `mix format`.

Hope that helps!


josevalim:

PR #1128, opened 2018-06-23
Fix a documentation bug, and add 3 reference links

smorin:
There is a bug referencing the wrong repo in one instance.  Added three references to make it easier for readers of the documentation.

### DIFF
--- a/getting-started/mix-otp/introduction-to-mix.markdown
+++ b/getting-started/mix-otp/introduction-to-mix.markdown
@@ -43,7 +43,8 @@ In this chapter, we will create our first project using Mix and explore differen
 >
 > If you have any questions or improvements to the guide, please reach discussion channels such as the [Elixir Forum](https://elixirforum.com) or the [issues tracker](https://github.com/elixir-lang/elixir-lang.github.com/issues). Your input is really important to help us guarantee the guides are accessible and up to date!
 >
-> The final code for this guide is in [this repository](https://github.com/elixir-lang/elixir-lang.github.com) and can be used as a reference.
+> The final code for the application built in this
+guide is in [this repository](https://github.com/josevalim/kv_umbrella) and can be used as a reference.
 
 > The Elixir guides are also available in EPUB format:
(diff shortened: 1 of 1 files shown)


getting-started/mix-otp/introduction-to-mix.markdown
```diff
@@ -33,17 +33,17 @@ In order to build our key-value application, we are going to use three main tool
...
+* ***[ExUnit](https://hexdocs.pm/ex_unit/)*** is a test-unit based framework that ships with Elixir;
 
 In this chapter, we will create our first project using Mix and explore different features in  <abbr title="Open Telecom Platform">OTP</abbr>, Mix and ExUnit as we go.
 
 > This guide requires Elixir v1.6.1 or later. You can check your Elixir version with `elixir --version` and install a more recent version if required by following the steps described in [the first chapter of the Getting Started guide](/install.html).
 >
 > If you have any questions or improvements to the guide, please reach discussion channels such as the [Elixir Forum](https://elixirforum.com) or the [issues tracker](https://github.com/elixir-lang/elixir-lang.github.com/issues). Your input is really important to help us guarantee the guides are accessible and up to date!
 >
-> The final code for this guide is in [this repository](https://github.com/josevalim/kv_umbrella) and can be used as a reference.
+> The final code for this guide is in [this repository](https://github.com/elixir-lang/elixir-lang.github.com) and can be used as a reference.
```
michalmuskala: I think the previous link was correct - it's supposed to be the source code for the thing you build throughout the guide, not the guide itself. 

Maybe we e should make the wording clearer?

josevalim:
Good call. We should say “the final code for the application built on this
guide” or similar. We probably don’t need to link to the guide itself as we
do so at the end of every page.
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

smorin:
Updated as requested

josevalim:

PR #1129, opened 2018-06-25
Use `console` language for command line examples in guides

sonicdoe:
`bash` denotes a shell script whereas `console` denotes a shell session.

Basic examples don’t look very different (before first, after second):

Before
After

One benefit is that Bash keywords are no longer highlighted (`for` in this case):

Before
After

The only downside I could find is that comments in the output are no longer highlighted:

Before
After

However, for the sake of consistency, I’ve changed them to `console` as well.


josevalim:

PR #1131, opened 2018-06-25
Fix guide reintroducing `Task.start_link/1`

sonicdoe:
`start_link/3` was changed to `start_link/1` in https://github.com/elixir-lang/elixir-lang.github.com/commit/cb0987997be2b463399510c6b690f7ed52913168. Therefore, it doesn’t really make sense to refer to `start_link/3` and to re-explain that `start_link/1` receives an anonymous function.


josevalim:

PR #1133, opened 2018-06-26
Add example of using iex -S mix compile

smorin:
Add example of using iex -S mix compile

### DIFF
--- a/getting-started/introduction.markdown
+++ b/getting-started/introduction.markdown
@@ -55,7 +55,7 @@ It seems we are ready to go! We will use the interactive shell quite a lot in th
 
 > Note: if you are on Windows, you can also try `iex.bat --werl` which may provide a better experience depending on which console you are using.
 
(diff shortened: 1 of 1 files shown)


smorin:
Usually true but as I recently discovered if the default mix task is replaced may not be the case.

josevalim:
`mix compile` won't load dependent modules. It will load them only if they happened to be compiled at the moment. But if you run `iex -S mix compile` twice, then nothing should be available the second time. If the default mix tasks is replaced and you want to run your app, then `iex -S mix run` is the way to go.

Btw, we talk about this on the supervisor chapter: https://elixir-lang.org/getting-started/mix-otp/supervisor-and-application.html (search for `iex -S mix run`).

smorin:
Thanks see the references there very helpful.

getting-started/introduction.markdown
```diff
@@ -55,6 +55,8 @@ It seems we are ready to go! We will use the interactive shell quite a lot in th
 
 > Note: if you are on Windows, you can also try `iex.bat --werl` which may provide a better experience depending on which console you are using.
 
+> Note: if you want to run `iex` with the project and dependent modules automatically loaded in interactive mode run `iex -S mix run`. See [Supervisor and application](/getting-started/mix-otp/supervisor-and-application.html) for more details.
```
josevalim: I am afraid this will be confusing for developers getting started because at this point they have no idea what `mix` is. Should they run this command? Does it matter? What are "dependent modules"? What is interactive mode? Remember this is the first chapter of the getting started.

smorin: Okay makes sense.  I'll move this to the next section that talks about IEX

josevalim:

PR #1136, opened 2018-07-02
binaries-strings-and-char-lists - fix misplaced paragraph

EliteTK:
commit 786d32ecc99ece297e4171e914e695408a969fde contributed information
regarding binary and char list concatenation but it resulted in a
paragraph referring to a previous example being pushed into the wrong
place.

This patch fixes this issue.


josevalim:

PR #1137, opened 2018-07-06
Update structs.markdown

Gerard-at-Streamit:
To get the keys of the struct, 'john' is referenced, where 'kurt' would be more logical in the context of the code snippet


josevalim:
Great catch. We are going too wild with the names here, so we should either stick wit John / Jane Doe or Bob and Alice, which are the most common examples found in textbooks. I have amended it, thank you!

josevalim:

PR #1141, opened 2018-07-11
Clarify sentence on configuration limitations in umbrella apps

Stratus3D:
It was unclear what was meant by different configuration for different applications. It makes more sense to say "where you want to use different configurations in each application for the same dependency" as it's clear the different configuration is for the same dependency.


whatyouhide:
Thanks @Stratus3D! 💟

PR #1146, opened 2018-07-24
Make Dynamic Dispatch example executable.

schmalz:
To conform to the typespec in the Parser behaviour, the two parse functions in the implementations in JSONParser and YAMLParser should return a tuple. This makes the example executable as is.


josevalim:

PR #1148, opened 2018-07-25
Update Learning page with Programming Elixir 1.6 and Adopting Elixir

lackac:
I also wanted to add the new edition of Elixir in Action, but unfortunately that is still in MEAP (I thought it was planned for a June release).

I thought about where to insert Adopting Elixir and decided on the third place because it's a logical next step after the first two books. :) It's a bit more advanced content, but important enough to include it towards the top of the list. Let me know or just move it if you prefer a different order.


josevalim:
Just to check, did you run something like [ImageOptim](https://imageoptim.com/) on the images? :)

lackac:
@josevalim no I haven't, but they are directly from PragProg's website, and I assume they have. :) I did check the file size of the the Programming Elixir 1.6 image and it was about third of the previous 1.3 image.

josevalim:

PR #1151, opened 2018-08-01
Show string-building anonymous function without parens

nathanl:
Functions like these are used in example factories for thoughtbot/ex_machina


michalmuskala:
Thank you! :heart:

PR #1162, opened 2018-08-20
Clarify ambiguity in race conditions

pik694:
Former text could give misleading impression on how to resolve interprocess communication problems.

I am also not sure whether the test "removes bucket on crash" behaves deterministically.
Agent.stop/2 is a sync method, which means that by the time the shell receives control the bucket is down; but can we be sure that the registry receives the message about bucket being `:DOWN` before it receives the message `:create`?  If it was deterministic messages should be sent in the following order:
1. Shell sends `:shutdown` to the bucket;
2. VM sends `:DOWN` message to the registry;
3. Shell sends `:create` message to registry.
Above order ensures that the test is deterministic.

I have not found the proper documentation to support my claim, but can't the order be as following?
1. Shell sends `:shutdown` to the bucket;
2/3. VM sends `:DOWN` message to the registry;
2/3. Shell sends `:create` message to registry.
Above order means that we are not sure which message reaches registry's inbox first - `:create:` or `:DOWN`.

I think it requires further investigation or a clear answer.

### DIFF
--- a/getting-started/mix-otp/ets.markdown
+++ b/getting-started/mix-otp/ets.markdown
@@ -209,7 +209,7 @@ Let's run the tests once again. This time though, we will pass the `--trace` opt
 $ mix test --trace
 ```
 
-The `--trace` option is useful when your tests are deadlocking or there are race conditions, as it runs all tests synchronously (`async: true` has no effect) and shows detailed information about each test. This time we should be down to one or two intermittent failures:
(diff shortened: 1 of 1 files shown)


josevalim:
All DOWN messages are sent at the same time. :) With this in mind, can you please send a sketch of a PR to improve this? Thanks!

josevalim:

PR #1164, opened 2018-08-31
Update task-and-gen-tcp.markdown

ciceropablo:


josevalim:
Although both ways are correct, the one you propose is more common, so :+1:.

josevalim:

ciceropablo:
Thanks, @josevalim.

PR #1165, opened 2018-09-01
rephrasing to avoid mis-matched pronouns.

annag42:
changed from a phrase using the mis-matched pronouns "one" and "their" to a rephrasing without any such pronouns.


josevalim:

PR #1166, opened 2018-09-01
change pronoun to match number, of "the community"

annag42:
changed from the possessive pronoun "their", in "their own tools", referring to the tools of "the community", to the possessive pronoun "its", since "the community" is a single entity. 
if "the community" is to be considered as being plural, please consider changing from "has also created" to "have also created".


josevalim:

PR #1167, opened 2018-09-05

ciceropablo:


getting-started/mix-otp/docs-tests-and-with.markdown
```diff
@@ -61,7 +61,7 @@ Doctests are specified by an indentation of four spaces followed by the `iex>` p
 
 Also, note that we started the documentation string using `@doc ~S"""`. The `~S` prevents the `\r\n` characters from being converted to a carriage return and line feed until they are evaluated in the test.
 
-To run our doctests, we'll create a file at `test/kv_server/command_test.exs` and call `doctest KVServer.Command` in the test case:
+To run our doctests, we'll create a file at `test/kv/command_test.exs` and call `doctest KVServer.Command` in the test case:
```
josevalim: Hi @ciceropablo!

Thanks for the PR. I believe the previous version was correct though, since the module is `KVServer`.

Thoughts?

PR #1169, opened 2018-09-05
Fix wrong stable version leading to wrong download links

Nicd:
Link pointed to 1.6 as latest stable, but latest stable is 1.7.

**NOTE:** I am unable to test this commit, because I cannot install the development tools. The gems include nokogiri, which cannot be installed because of this problem: https://github.com/sparklemotion/nokogiri/issues/1772

But I think this should work. :P


josevalim:

PR #1171, opened 2018-09-08
Add link to Collectable protocol documentation

link2xt:


link2xt:
This is the first time it is mentioned.

By the way, what is the policy on the links to documentation? Should it be only when something is mentioned the first time or does it make sense to add as many links as possible?

link2xt:
@eksperimental Fixed it

eksperimental:
I don't think there's such a policy but linking the first time it is mentioned it's good enough for me

eksperimental:
Thank you @link2xt

PR #1178, opened 2018-09-25
Wrap by backquote

deraru:
Hello,

We are now translating this guide into Japanese on the https://github.com/elixir-lang-jp/elixir-lang.github.com repository.
Then I found some non backquoted word at "try, catch, and rescue" section.
Could you review this?

Thanks 😄


josevalim:

PR #1179, opened 2018-09-28
Add link to VSCode extension

vassilevsky:
CC @fr1zle


fr1zle:
The link "Visual Studio Elixir" already links to the extension in question.

vassilevsky:
Should I change the link text then? Not sure if this extension will work in the big Visual Studio.

Issue #1084, opened 2018-01-30
Cannot complete Dynamic Supervisor guide


sentyaev:
On the Dynamic supervisors page there is a section The bucket supervisor.
I'm very new to Elixir and code in this section looks incomplete for me.
This is the code block:

> Let’s define a DynamicSupervisor and give it a name of KV.BucketSupervisor in lib/kv/application.ex as follows:
```
  def init(:ok) do
    children = [
      {KV.Registry, name: KV.Registry},
      {DynamicSupervisor, name: KV.BucketSupervisor, strategy: :one_for_one}
    ]

    Supervisor.init(children, strategy: :one_for_one)
  end
```

I've created a file `application.ex` like this:
```
defmodule KV.BucketSupervisor do
  def init(:ok) do
    children = [
      {KV.Registry, name: KV.Registry},
      {DynamicSupervisor, name: KV.BucketSupervisor, strategy: :one_for_one}
    ]

    Supervisor.init(children, strategy: :one_for_one)
  end
end
```

After I execute this command:
```
iex> {:ok, bucket} = DynamicSupervisor.start_child(KV.BucketSupervisor, KV.Bucket)
```

It fails with the following error:
```
** (exit) exited in: GenServer.call(KV.BucketSupervisor, {:start_child, {{KV.Bucket, :start_link, [[]]}, :permanent, 5000, :worker, [KV.Bucket]}}, :infinity)
    ** (EXIT) no process: the process is not alive or there's no process currently associated with the given name, possibly because its application isn't started
    (elixir) lib/gen_server.ex:821: GenServer.call/3
```

Not sure is my fault or documentation is incomplete/or not clear.

josevalim:
Great catch @sentyaev! It was supposed to say:

> Let’s define a DynamicSupervisor and give it a name of KV.BucketSupervisor in **lib/kv/supervisor.ex** as follows:

Note we were pointing you to application.ex but you were supposed to change the existing lib/kv/supervisor.ex file. If you remove the application.ex file and change supervisor.ex accordingly you should be good to go!

PR #1180, opened 2018-10-04
Refine DynamicSupervisor instructions in Mix/OTP Guide

avonengel:
The previous instructions tripped me up a little: First, I created a new module `KV.BucketSupervisor` with the `init` function from the guide. My mind kept auto-correcting "in `lib/kv/supervisor.ex`" to "in `lib/kv/bucketsupervisor.ex`" ;)

After I found #1084, I replaced the whole content of the `KV.Supervisor` module, due to the instructions saying
> Note this time we didn’t have to define a separate module that invokes use DynamicSupervisor

this results in `(UndefinedFunctionError) function KV.Supervisor.start_link/1 is undefined or private`.

Then I finally got it right ;)


josevalim:

PR #1187, opened 2018-11-04
fixes, warning: variable "a" is unused.

MichaelDimmitt:
Additionally, because the variable is called before and after the macro.

Tracing how the value changes is easier to implement. 
```elixir
  def go do
    require Hygiene
    a = 13
    a
    Hygiene.interference
    a
  end
```
Can easily be modified to log the values:
```elixir
  def go do
    require Hygiene
    a = 13
    a |> IO.inspect 
    Hygiene.interference
    a |> IO.inspect 
  end
```
Without changing the end result.

This page helped a lot with my macro learning 👍 
The warning was the semi confusing part because it brings uncertainty to whether the original value is unused or the overriding value is unused. If this warning was intentional including the reasoning for the warning could be an alternative solution for the problem.

### DIFF
--- a/getting-started/meta/macros.markdown
+++ b/getting-started/meta/macros.markdown
@@ -126,7 +126,6 @@ defmodule HygieneTest do
   def go do
     require Hygiene
     a = 13
-    a
     Hygiene.no_interference
     a
   end
@@ -149,7 +148,6 @@ defmodule HygieneTest do
   def go do
     require Hygiene
     a = 13
-    a
     Hygiene.interference
     a
   end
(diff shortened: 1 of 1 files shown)


josevalim:
Good catch @MichaelDimmitt!

Instead of changing the code, and making the example we want to show harder to understand, should we change the text that, as a result of the `no_interference`, we even get unused variable warnings?

MichaelDimmitt:
@josevalim, sounds like a good idea. Want me to attempt explaining that info?

josevalim:
Yes, that would be lovely, thank you!

josevalim:

PR #1195, opened 2018-12-10
Update ETS markdown

andreolvr:
`GenServer` module doesn't provide `call/2` function. I checked the [documentation]( https://hexdocs.pm/elixir/search.html?q=GenServer.call) and the [source code](https://github.com/elixir-lang/elixir/blob/v1.7.4/lib/elixir/lib/gen_server.ex#L911) and didn't find any `GenServer.call/2` function.


michalmuskala:
The default means there are two functions `GenServer.call/2` and `GenServer.call/3` - they are both defined with that single definition.

But I agree a more canonical way is to refer to the maximum arity function in case of defaults.

PR #1196, opened 2018-12-14
Fix a statement about the heaviness of threads in other programming languages

brettcannon:
I think the initial intent was to write "not unlike", but the "not" got dropped.


michalmuskala:
The original sentence is correct. Threads are very heavyweight compared to Elixir/Erlang processes. Maybe we should reword this sentence a bit more to make that point more obvious?

brettcannon:
Yeah, some rewording might be good then because traditionally when comparing processes to threads the former is considered heavyweight and the latter is lightweight. So having Elixir want to make the point "and we're even lighter!" it probably needs to be a bit more explicit when mentioning both processes and threads in the same sentence.

brettcannon:
OK, I tried to come up with different wording. Please have a look.

josevalim:

PR #1198, opened 2018-12-14
Harmonize the wording of two sentences

brettcannon:
"a list of integers representing bytes or binaries" compared to "a list of characters or strings" in two adjacent paragraphs is slightly jarring. Changing both to "or" makes the comparison/transition easier to read.


michalmuskala:
Thank you! :heart:

PR #1199, opened 2018-12-21
Add a note about being in a defmodule to refer to a struct

brettcannon:
If you follow the tutorial in a `.ex` or `.exs` file, then referring to a struct leads to an unexpected error about it not being defined. (This has already come up on Stack Overflow: https://stackoverflow.com/questions/39576209/elixir-cannot-access-struct.)

### DIFF
--- a/getting-started/structs.markdown
+++ b/getting-started/structs.markdown
@@ -35,7 +35,7 @@ The keyword list used with `defstruct` defines what fields the struct will have 
 
 Structs take the name of the module they're defined in. In the example above, we defined a struct named `User`.
 
-We can now create `User` structs by using a syntax similar to the one used to create maps (although do note that you must be within a `defmodule` in a `.ex` or `.exs` file to refer to a struct, otherwise you will get an error saying `the struct was not yet defined`):
+We can now create `User` structs by using a syntax similar to the one used to create maps (if you have defined the struct in a separate file, you can compile the file inside IEx before proceeding by running `c "file.exs"`; be aware you may get an error saying `the struct was not yet defined` if you try the below example in a file directly due to when definitions are resolved):
 
 ```iex
 iex> %User{}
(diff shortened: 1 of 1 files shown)


getting-started/structs.markdown
```diff
@@ -35,7 +35,7 @@ The keyword list used with `defstruct` defines what fields the struct will have
 
 Structs take the name of the module they're defined in. In the example above, we defined a struct named `User`.
 
-We can now create `User` structs by using a syntax similar to the one used to create maps:
+We can now create `User` structs by using a syntax similar to the one used to create maps (although do note that you must be within a `defmodule` in a `.ex` or `.exs` file to refer to a struct, otherwise you will get an error saying `the struct was not yet defined`):
```
josevalim: Then maybe we should say:

> If you have defined the struct in IEx, then it should be immediately available. If you have defined it in a separate file, you can compile the file inside IEx before proceeding by running `c "file.exs"`.

In other words, you don't need to be within the file or within within IEx, but you need to make sure the struct definition is available.

Thoughts?

brettcannon: For me personally the issue was I was doing none of the tutorial in IEx but entirely in a scratch file, so knowing how to import in IEx would not have necessarily helped (but I also read ahead to learn how to make `IO.inspect` work, so maybe my use-case isn't _too_ important 😉 ).

josevalim: Your call. :) We can expand the current information but it is more of a matter of making sure the struct has been loaded into memory than having to put in the same file and so on.

brettcannon: Here's another shot at trying to merge your suggestion, José, while not having folks typing this out in a file get taken by surprise if they get the error message (rather than trying to talk them through why they get the error).

josevalim:

PR #1201, opened 2019-01-02
Add a semicolon, remove Oxford comma

ashton314:
Line 81 had an [Oxford comma](https://en.wikipedia.org/wiki/Serial_comma), but the rest of the document does not use Oxford commas.

The comma on line 112 would be better with a semicolon: it's delimiting two related clauses that *could* stand on their own and so should be connected with a semicolon rather than a comma.


josevalim:

PR #1204, opened 2019-01-04
Don't use assignment pattern matching in function parameters

brettcannon:
It doesn't seem to be covered or used anywhere in the entire Getting Started guide, so its sudden use is a bit jarring.

Also add some code literal markup in one place.


josevalim:

PR #1207, opened 2019-01-05
End all comments with a period for ETS page

brettcannon:


josevalim:

PR #1208, opened 2019-01-05
Consistently use periods at the end of comments for doc tests page

brettcannon:
Previously the file _mostly_ used periods, but not everywhere.


josevalim:

PR #1209, opened 2019-01-05
Add periods at the end of each comment of Distributed Tasks page

brettcannon:
Previously the use of periods was inconsistent.


josevalim:

PR #1210, opened 2019-01-05
End a comment in a period on the Dynamic Supervisor page

brettcannon:


josevalim:
Closing for the reasons mentioned. I think it is probably better to standardize on no periods at the last line!

PR #1211, opened 2019-01-05
Use proper punctation for comments on Agent page

brettcannon:


josevalim:
Closing for the reasons mentioned. I think it is probably better to standardize on no periods at the last line!

josevalim:
@brettcannon it is no noise at all. The contributions are definitely welcome!

To be fair, this is not documented at all, so we are going on whatever the generators generate as documentation. :D

PR #1212, opened 2019-01-05
Add periods to the end of comments on the DSL page

brettcannon:


josevalim:
Closing for the reasons mentioned in another issue. I think it is probably better to standardize on no periods at the last line! Thanks!

PR #1222, opened 2019-01-25
Fix arity of alias!/1

### LINKED ISSUE #1190: Macros guide: Wrong arity of alias!/1
https://github.com/elixir-lang/elixir-lang.github.com/edit/master/getting-started/meta/macros.markdown#L182

> Elixir provides similar mechanisms for imports and aliases too. This guarantees that a macro will behave as specified by its source module rather than conflicting with the target module where the macro is expanded. Hygiene can be bypassed under specific situations by using macros like `var!/2` and `alias!/2`, although one must be careful when using those as they directly change the user environment.

`alias!/2` might be [`alias!/1`](https://hexdocs.pm/elixir/Kernel.html#alias!/1).

cdfuller:
Fixes #1190


josevalim:

PR #1223, opened 2019-02-04
Add explanation about why atom can be useful

Marcos-Costa:
This simple text is primary focused in newcomers to the language (as myself). Which have low or none past experience using function programming. The text is an answer that I've found after researching on the web to understand why introduce a new type instead of comparing two strings.

### DIFF
--- a/getting-started/basic-types.markdown
+++ b/getting-started/basic-types.markdown
@@ -116,7 +116,9 @@ iex> :hello == :world
 false
 ```
 
-The booleans `true` and `false` are, in fact, atoms:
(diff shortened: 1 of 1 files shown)


eksperimental:
This is the very first lesson, and we haven't even stalked about strings.
And to be honest, it is a very specific use case the one mentioned there. I don't even know if I would include it somewhere it.

josevalim:
Thanks @Marcos-Costa! As @eksperimental said, this is too early to talk about this. I offer two suggestions:

1. We can talk about this in the pattern matching chapter. We can say it is common to use atoms on "tagged tuples" and explain why.

2. We can talk about this in the module documentation for the Atom module: https://github.com/elixir-lang/elixir/blob/v1.8.1/lib/elixir/lib/atom.ex#L3

Thoughts?

PR #1226, opened 2019-02-13
Add module attributes section

EskiMag:
I think `@for` and `@protocol` module attributes should be mentioned in the **Protocols** getting started guide to prevent confusion.

Related topic: https://elixirforum.com/t/using-module-in-protocol-implementation/20082


josevalim:
Hi @EskiMag! The getting started guide is not meant to be a complete reference. Instead what we do is to link to `defimpl` and `defprotocol` to provide a complete reference  which does provide the information you have mentioned. Thank you!

EskiMag:
Thanks, @josevalim. That's a valid point. I am just not sure if you meant that the links to `defimpl` and `defprotocol` will be added or are already somewhere. I was not able to find them here https://elixir-lang.org/getting-started/protocols.html

josevalim:
Yup, I am adding them now because I am refactoring things down a bit and improve the module docs for Protocols in general. :D

PR #1234, opened 2019-02-28
Updated about macOS Mojave Dark mode problem

vaibhav-kaushal:
Erlang observer GUI is inconsistent (tab bar becomes white over white and sometimes the graph under Applications tab is not shown) with the macOS Mojave Dark Mode. I think better language could be used though. 

Screenshot : 

screenshot 2019-02-28 at 8 41 54 am


josevalim:
Hi @vaibhav-kaushal, sorry for the delay. I have verified this on Erlang/OTP 22 and it seems to work without problems now. Thanks for the report!

PR #1236, opened 2019-03-05
Bring back outline styles

kennethlarsen:
This PR brings back styling for outline when an element is focused. This is tremendously important for accessibility: http://www.outlinenone.com/


josevalim:

PR #1238, opened 2019-03-13
fix wrong url for opensuse distro obs://devel:languages:erlang/ url

MalloZup:
# previous behaviour:
repo was picked-up in wrong way because the Url logic.

So an user could endup to install wrong pkgs from elixir/erlang, which can led to bugs of incompatibility vs elixir and erlang version

# new behaviour
when specified obs://devel:languages:erlang/ url, then zypper does look for current right distro version.
In this way each openSUSE distro has the right erlang/elixir pkgs.

### DIFF
--- a/install.markdown
+++ b/install.markdown
@@ -32,8 +32,9 @@ If your distribution contains an old Elixir/Erlang version, see the sections bel
   * Arch Linux (Community repo)
     * Run: `pacman -S elixir`
   * openSUSE (and SLES)
-    * Add Erlang devel repo: `zypper ar -f obs://devel:languages:erlang/ erlang`
+    * add Elixir/Erlang repo: `zypper ar -f obs://devel:languages:erlang/ Elixir-Factory`
     * Run: `zypper in elixir`
+    * optional: if you want to use the latest Erlang, you can use this repo: `zypper ar -f  obs://devel:languages:erlang:Factory Erlang-Factory`
   * Gentoo
     * Run: `emerge --ask dev-lang/elixir`
   * GNU Guix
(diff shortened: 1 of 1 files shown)


MalloZup:
just some more info about this PR:

when using a distro like opensuse Leap15, without this change an USER can pick up the wrong erlang/elixir which can led to uncompatibility issues, eg. crashes like

```
{"init terminating in do_boot",{undef,[{elixir,start_cli,[],[]},{init,start_em,1,[]},{init,do_boot,3,[]}]}}
init terminating in do_boot ({undef,[{elixir,start_cli,[],[]},{init,start_em,1,[]},{init,do_boot,3,[]}]})
```


thx for your review :surfing_man: :sunflower:

Gsantomaggio:
I would add also the URL for the last erlang version:
```
zypper ar -f  obs://devel:languages:erlang:Factory SUSE-Erlang-Factory
``` 
and I'd modify 
```
zypper ar -f obs://devel:languages:erlang/ SUSE-Elixir-Factory 
```

MalloZup:
thx @Gsantomaggio  updated. I removed the `SUSE` because it can be to much Distro specific, in case of SLES or openSUSE. So imho is better to have without prefix. 

updated in latest commit.

I have also discovered that the web-page  from `erlang` doesn't have `openSUSE`.

https://www.erlang.org/downloads
also the 

https://www.erlang-solutions.com/resources/download.html 

doesn't have any mention of openSUSE. We could afaik send a PR to this docs if they are hosted in GitHub.

Gsantomaggio:
@MalloZup 
> https://www.erlang.org/downloads

provides only the windows packages

>https://www.erlang-solutions.com/resources/download.html

The erlang-solutions packages are not open-source

MalloZup:
thx @Gsantomaggio for infos , didn't know it :sun_with_face:

josevalim:

PR #1239, opened 2019-03-24
The bucket supervisor; Improved instructions

fidenz-chim:
Instructions were ambiguous, so wanted those to be specific. Included what the user should do with the DynamicSupervisor and where.


josevalim:

PR #1241, opened 2019-03-28
Specific instructions on how to run test

fidenz-chim:
Instructions to specify default $PORT in application.ex and how to run test

### DIFF
--- a/getting-started/mix-otp/docs-tests-and-with.markdown
+++ b/getting-started/mix-otp/docs-tests-and-with.markdown
@@ -70,7 +70,13 @@ defmodule KVServer.CommandTest do
 end
 ```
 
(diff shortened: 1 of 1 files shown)


josevalim:
Hi @fidenz-chim!

We have already changed the code to no longer raise and always use 4040 [in this commit](https://github.com/elixir-lang/elixir-lang.github.com/commit/616d78c7118587594865b24ae23451908f98991b). So why would this change also be necessary? Thank you.

fidenz-chim:
@josevalim  My mistake, I have changed my code in between testing different scenarios. Sorry about any inconveniences.

josevalim:
No problem, thanks for contributing!

PR #1243, opened 2019-04-03
Want to include the resources to learn elixir

diveshjain25:


josevalim:
We'd prefer to not link to intermediate resources but if any of those learning materials would like to add links to our learning resources page, we would welcome it! Thanks for the PR @diveshjain25!

PR #1246, opened 2019-04-07
Adding note about headless vs gui-enabled erlang

PaulOstazeski:
On at least some linux distributions, the headless version of erlang is installed as the default dependency for elixir. This results in an error when trying to start the observer or other gui tools.

```elixir
iex> :observer.start
{:error,
 {{:load_driver, 'No driver found'},
  [...]
}}
```

### DIFF
--- a/getting-started/mix-otp/dynamic-supervisor.markdown
+++ b/getting-started/mix-otp/dynamic-supervisor.markdown
@@ -170,9 +170,9 @@ Now that we have defined our supervision tree, it is a great opportunity to intr
 iex> :observer.start
 ```
 
-> Note: If you receive an `{:error, ...}` tuple similar to "No driver found" instead of seeing the observer window, here's what's happened: some package managers default to installing a minimized Erlang without WX bindings for GUI support when you install Elixir. If this affects you, you can either replace the headless Erlang with a more complete package (look for packages named `erlang` vs `erlang-nox` on debian/ubuntu/arch), or you can continue using your current version without being able to use GUI tools.
-
 A GUI should pop-up containing all sorts of information about our system, from general statistics to load charts as well as a list of all running processes and applications.
+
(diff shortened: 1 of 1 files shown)


getting-started/mix-otp/dynamic-supervisor.markdown
```diff
@@ -170,6 +170,8 @@ Now that we have defined our supervision tree, it is a great opportunity to intr
 iex> :observer.start
 ```
 
+> Note: If you receive an `{:error...` tuple that mentions 'No driver found', ensure that you have installed `erlang` and not the headless `erlang-nox` package from your package manager. You can continue using the headless version, but will have no GUI tools.
```
whatyouhide:
```suggestion
> Note: If you receive an `{:error, ...}` tuple that mentions "No driver found", ensure that you have installed `erlang` and not the headless `erlang-nox` package from your package manager. You can continue using the headless version, but you will have no GUI tools.
```

getting-started/mix-otp/dynamic-supervisor.markdown
```diff
@@ -170,6 +170,8 @@ Now that we have defined our supervision tree, it is a great opportunity to intr
 iex> :observer.start
 ```
 
+> Note: If you receive an `{:error, ...}` tuple that mentions "No driver found", ensure that you have installed `erlang` and not the headless `erlang-nox` package from your package manager. You can continue using the headless version, but you will have no GUI tools.
```
josevalim: The erlang-nox is specific to certain packager managers but others may have the same issue with other names. My suggestion is something like this:

```suggestion
> Note: Some distributions and package managers do not install Erlang's complete standard library when you install Erlang. Therefore, if you receive an `{:error, ...}` tuple that mentions "No driver found" or similar, make sure that you have installed the complete version of Erlang or at least include its WX bindings. If you prefer, you can continue using your current version, but you will have no GUI tools.
```

Thoughts?

josevalim:

PR #1247, opened 2019-04-12
Clarify that cond conditions can return any truthy value, not just `true`

Stratus3D:
Ran into an issue today, and when reading the docs I got the impression cond
looked specifically for `true`, and not just any truthy value.

### DIFF
--- a/getting-started/case-cond-and-if.markdown
+++ b/getting-started/case-cond-and-if.markdown
@@ -100,7 +100,7 @@ iex> f2 = fn
 
 ## `cond`
 
-`case` is useful when you need to match against different values. However, in many circumstances, we want to check different conditions and find the first one that evaluates to true. In such cases, one may use `cond`:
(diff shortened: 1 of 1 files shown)


Stratus3D:
I'm not sure if the change to the first sentence in this PR is strictly required.

josevalim:
Hi @Stratus3D!

We don't use the concept of truthy and falsy value anywhere else in the docs. So if we want to use it here, we would need to explicitly declare what it is before. We probably define the concept of `nil` and `false` being considered false at some point, so that would be a good opportunity to introduce said concepts.

josevalim:
Ping.

josevalim:
Ping.

Stratus3D:
Sorry, I've not forgotten about this but I haven't had a chance to think about how this concept should be introduced. Should I add a paragraph on falsey and truthy values under the `cond` header?

josevalim:
Perhaps? Another idea is to say any clause that doesn't return `nil` or `false`. And avoid the general truthy/falsy discussion for now.

josevalim:

PR #1255, opened 2019-05-21
Clarify boolean operator rule

jeansberg:
The original text makes it sound like you need a 'true' or 'false' literal on the left side.


josevalim:

PR #1256, opened 2019-05-23
add keyboard navigation into getting started guide

ignatiusreza:
This PR tries to add left/right keyboard shortcut to the getting started section, which will activate `previous` and `next` navigation respectively, since I find it easier to navigate with keyboard shortcut than having to reach for the mouse..

Hopefully it will help.. :tada:

### DIFF
--- a/_layouts/getting-started.html
+++ b/_layouts/getting-started.html
@@ -54,7 +54,7 @@ section: getting-started
               <a class="page-numbers spec" href="#container">Top</a>
             {% unless forloop.last %}
               {% assign next = guide.pages[forloop.index] %}
-              <a href="{{guide.dir}}{{next.slug}}.html" class="previous page-numbers spec" title="{{next.title | escape}}">Next &rarr;</a>
+              <a href="{{guide.dir}}{{next.slug}}.html" class="next page-numbers spec" title="{{next.title | escape}}">Next &rarr;</a>
             {% endunless %}
           </div>
         {% endif %}
--- a/_includes/bottom.html
+++ b/_includes/bottom.html
@@ -12,6 +12,28 @@
   <script src="/js/toc/toc.js" type="text/javascript" charset="utf-8"></script>
   <script type="text/javascript">
     $(document).ready(function() {
+      $(document).on('keyup', function(e) {
+        if (e.altKey || e.ctrlKey || e.metaKey || e.shiftKey) return;
+
+        switch(e.which) {
+          case 37: // left arrow
+            goToPage('previous');
+          break;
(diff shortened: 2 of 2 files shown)


_layouts/getting-started.html
```diff
@@ -54,7 +54,7 @@ <h3 class="widget-title">{{guide.title}}</h3>
               <a class="page-numbers spec" href="#container">Top</a>
             {% unless forloop.last %}
               {% assign next = guide.pages[forloop.index] %}
-              <a href="{{guide.dir}}{{next.slug}}.html" class="previous page-numbers spec" title="{{next.title | escape}}">Next &rarr;</a>
+              <a href="{{guide.dir}}{{next.slug}}.html" class="next page-numbers spec" title="{{next.title | escape}}">Next &rarr;</a>
```
ignatiusreza: I looked around and couldn't find any places where `.previous` is being used by either css or js.. so I changed the class here from `previous` to `next`, so that we can easily differentiate between the two links..

josevalim:
Hi @ignatiusreza!

I am personally not a big fan of using the arrows for navigation. Is there a link or references indicating that using arrows for this kind of material is a good practice UX/Accessibility wise? Thanks!

ignatiusreza:
I encountered several tutorial (book) type content with similar keyboard shortcut in the past, but I'm not sure if there's references or study on it actually.. :thinking: 

If arrows key is not a good candidate, would it be more preferable to use vim-style (j & k) keyboard shortcut? as an example for this, gmail use j & k to navigate between emails.. wdyt?

josevalim:
> I encountered several tutorial (book) type content with similar keyboard shortcut in the past

I see. Do you have any examples? Most times I end up triggering those features by accident, when I was not expecting them... but maybe I am just getting old and grumpy :D

ignatiusreza:
hahaha, you're by no means old or grumpy :laughing: 

I don't have one on the top of my head, but i'll try to look around and will report back :+1:

ignatiusreza:
sorry, i tried to look around, but no longer could find one with similar navigation approach.. last time i found them was roughly 2 years ago, when i was reading a lot of books/tutorial about webpack, react, redux and related material.. but, since then, most of the official material has been rebuilt, and community ones gets outdated and replaced by newer ones..

i'll close this for now, until I can find more references..

ignatiusreza:
sorry from bringing an old topic back up, but i've found a couple examples:

- [elm guide book](https://guide.elm-lang.org/)
- [github search results page](https://github.com/elixir-lang/elixir-lang.github.com/search?q=example)

PR #1258, opened 2019-05-23
Added a link to the Code Elixir LDN conference to important links

gausby:
I placed it above the link to ElixirConf US as the Code Elixir LDN conference will take place sooner than ElixirConf.

Also, the image graphic is included in the anchor tag such that the link is only present once, which should be good for both people navigating the site using keyboard and screen readers.


josevalim:

PR #1259, opened 2019-05-23
Add a Nix setup allowing Nix users to invoke a Ruby/Jekyll setup

gausby:
Note: it won't hurt my feelings if this PR is rejected. I just set it up on my machine because I wanted to make some changes to the site. I know people running NixOS and NixPkg is very rare in the grand scale of things. I've made this pull request because I spent 10 minutes setting it up, and it might be useful to others.

The users would have to run the `nix-setup-or-initialize.sh` file, which will fetch Ruby and the dependencies from the Nix store and make them available locally. When that is complete it is possible to run the Jekyll setup by running `nix-shell` in the project directory; the changes made to the site should get picked up on site running locally on port 4000.

Not everyone in the world is using Nix (yet), but this makes it very easy to get started, and the Nix users will not have to have Ruby or Jekyll installed globally on their systems, only for the Nix session.

I have ignored the `.bundle` folder in the `.gitignore` as well; I do not know if this is the way to go.

The setup is based on the following blog post by Nathan Bijnens:

  - https://nathan.gs/2019/04/19/using-jekyll-and-nix-to-blog/

I would imagine a note about this in the readme would be needed before this PR can be accepted.


josevalim:
Yes, I think this is very specific but, by at least having the PR, it is documented for others who may be interested in the future. Thanks @gausby!

PR #1261, opened 2019-06-01
Add subreddit links.

alexrp:
Now that I've actually found the time to get the subreddit back in [working order](https://www.reddit.com/r/elixir/comments/bvqzc8/an_update_on_the_state_of_this_subreddit/), I thought it would make sense to add it to the list of community links.

I'm not sure how the list of links is supposed to be sorted. I went with the logic that since a subreddit is essentially a forum, the link would go after the official Elixir Forum. I have no particular preference though.

### DIFF
--- a/_includes/important-links.html
+++ b/_includes/important-links.html
@@ -23,6 +23,7 @@
   <h3 class="widget-title">Join the Community</h3>
   <ul>
     <li><a class="spec" href="http://elixirforum.com">Elixir Forum</a></li>
+    <li><a class="spec" href="https://www.reddit.com/r/elixir">Elixir on Reddit</a></li>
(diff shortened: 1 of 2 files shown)


josevalim:
Hi @alexrp! Thanks for taking care of the subreddit. The Elixir subreddit is probably the forum that I read the most of all and it will be nice to see a larger moderation team.

However, I am a bit worried about linking newcomers to the subreddit. While it doesn't happen in r/elixir, in Reddit it is unfortunately very easy to go to other subcommunities, even ones related to programming, that can be quite hostile, and that can be a poor experience to a newcomer.

In any case, you can link to the subreddit from the [Wiki](https://github.com/elixir-lang/elixir/wiki) alongside other resources.

PR #1263, opened 2019-06-04
File.close terminates its process.

mfrasca:
I think this is interesting to notice explicitly.

### DIFF
--- a/getting-started/io-and-the-file-system.markdown
+++ b/getting-started/io-and-the-file-system.markdown
@@ -45,9 +45,16 @@ iex> File.close(file)
 :ok
 iex> File.read("hello")
 {:ok, "world"}
+iex> IO.binwrite(file, "świat")
+{:error, :terminated}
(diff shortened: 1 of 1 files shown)


mfrasca:
I see that you do mention it a bit below in the text.  maybe a forward reference?

josevalim:
Oh, good find. If we are already talk about it, perhaps we can move the discussion down. Can you please update?

josevalim:
I haven't reviewed this chapter in a while, so I decided to remove the more advanced parts too and focus on the important bits for now. Thank you! 

mfrasca:
I found those "more advanced parts" quite interesting, I mean, there's so many ways to be a beginner, and in-depth materials can be eye openers.  it was already relegated into a "you don't want to know this" section.  anyhow, I hope you can use it somewhere else.  and I'll be very glad to know where that somewhere will be.  ciao, M

PR #1264, opened 2019-06-05
I had totally missed the text in the note

mfrasca:
would I have not missed it if it was more prominent, like I'm suggesting here?  I had reached the module attributes page, and finally realized there was a help function, so I said to myself let's put it somewhere in the text, where it should be introduced, and guess what I intended to put it just only one paragraph above the spot I had missed.


josevalim:

PR #1265, opened 2019-06-05
mention mixing both implicit and explicit defaults

mfrasca:
I'm not sure if this is a feature or a bug, but if a feature, it does no harm documenting it.


josevalim:

PR #1266, opened 2019-06-05
*anonymous* functions first class citizens

mfrasca:
coming from Python, this assertion had me confused, until I reached the `&` operator for named functions in "chapter 8".
I think it's better to first show anonymous functions by themselves, not matched to a variable, then to do that match, and to avoid the contradiction in terms 'named anonymous'.

### DIFF
--- a/getting-started/basic-types.markdown
+++ b/getting-started/basic-types.markdown
@@ -240,9 +240,9 @@ iex> is_function(add, 1)
 false
 ```
 
-Parenthesised arguments after the anonymous function indicate that we want the function to be evaluated, not just its definition returned. By contrast, Elixir attempts evaluation of named functions, even in absence of the trailing explicit `()` function call syntax. Matching an anonymous function to a variable does not promote it to the 'named' function status. We will explore named functions when dealing with [Modules and Functions](/getting-started/modules-and-functions.html), since named functions can only be defined within a module.
+Parenthesised arguments after the anonymous function indicate that we want the function to be evaluated, not just its definition returned.  Note that a dot (`.`) between the variable and parentheses is required to invoke an anonymous function. The dot ensures there is no ambiguity between calling the anonymous function matched to a variable `add` and a named function `add/2`. In this sense, Elixir makes a clear distinction between anonymous functions and named functions.
 
-Note that a dot (`.`) between the variable and parentheses is required to invoke an anonymous function. The dot ensures there is no ambiguity between calling the anonymous function matched to a variable `add` and a named function `add/2`. In this sense, Elixir makes a clear distinction between anonymous functions and named functions.
+We will explore named functions when dealing with [Modules and Functions](/getting-started/modules-and-functions.html), since named functions can only be defined within a module.
 
 Anonymous functions are closures and as such they can access variables that are in scope when the function is defined. Let's define a new anonymous function that uses the `add` anonymous function we have previously defined:
 
--- a/getting-started/basic-types.markdown
+++ b/getting-started/basic-types.markdown
@@ -223,7 +223,7 @@ iex> is_function(fn a, b -> a + b end)
 true
 ```
 
-Anonymous functions are "first class citizens" in Elixir, meaning they can be matched to variables, and passed as arguments to other functions in the same way as integers and strings. In the example above, we have passed an anonymous function definition to the `is_function/1` function which correctly returned `true`. In the example below, we match the anonymous function to a variable, evaluate it, and check the arity of the function by calling `is_function/2`.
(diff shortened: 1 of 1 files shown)


getting-started/basic-types.markdown
```diff
@@ -215,12 +215,23 @@ iex> String.upcase("hellö")
...
 #Function<12.71889879/2 in :erl_eval.expr/5>
-iex> add.(1, 2)
+iex> (fn a, b -> a + b end).(1, 2)
 3
-iex> is_function(add)
+iex> is_function(fn a, b -> a + b end)
 true
+```
+
+Anonymous functions are "first class citizens" in Elixir, meaning they can be matched to variables, and passed as arguments to other functions in the same way as integers and strings. In the example above, we have passed an anonymous function definition to the `is_function/1` function which correctly returned `true`. In the example below, we match the anonymous function to a variable, evaluate it, and check the arity of the function by calling `is_function/2`.
```
josevalim:
```suggestion
Anonymous functions are "first class citizens" in Elixir, meaning they can be assigned to variables, and passed as arguments to other functions in the same way as integers and strings. In the example above, we have passed an anonymous function definition to the `is_function/1` function which correctly returned `true`. Let's assign it to a variable next:
```

getting-started/basic-types.markdown
```diff
@@ -229,9 +240,9 @@ iex> is_function(add, 1)
 false
 ```
 
-Functions are "first class citizens" in Elixir meaning they can be passed as arguments to other functions in the same way as integers and strings. In the example, we have passed the function in the variable `add` to the `is_function/1` function which correctly returned `true`. We can also check the arity of the function by calling `is_function/2`.
+Parenthesised arguments after the anonymous function indicate that we want the function to be evaluated, not just its definition returned. By contrast, Elixir attempts evaluation of named functions, even in absence of the trailing explicit `()` function call syntax. Matching an anonymous function to a variable does not promote it to the 'named' function status. We will explore named functions when dealing with [Modules and Functions](/getting-started/modules-and-functions.html), since named functions can only be defined within a module.
```
josevalim: I generally don't like to talk about what something is not, because unless you come from a background where the negative affirmation exists, then it is more likely to make things confusing. Therefore I would remove this paragraph. And move the "We will explore..." sentence to the end of the next paragraph, as it was!

getting-started/basic-types.markdown
mfrasca: I'm curious.  why did you correct 'match' with 'assign'?  I'm sure we mean the same, but the 'match' term sounds more formally correct to me.

josevalim:
I have added two comments, thank you!

getting-started/basic-types.markdown
mfrasca: in the above example, I included the 'anonymous definition returned', this is why I added the discussion in the first sentence (Parenthesised…returned).
you consider the two following sentences (By contrast…syntax) and (Matching…status) confusing?  I can remove them from here.  I would like to mention the fact, where we go on explaining named functions.

josevalim: The named function syntax without parens that you refer to is "deprecated", users will get a warning and in the future it will straight fail, so I wouldn't worry about it.

But to me, the main issue with the section is that it is assuming the user will confuse variable assignment with named functions, but they may not. And simply mentioning this can be confusing (or make them confused while they were not before).

mfrasca: that's good to hear.  will it —some time in the future— mean that you can mention a named function `foo` without Elixir automatically looking for evaluating `foo/0`?  it would also make the `&` syntax superfluous, isn't it?

as for this edit, I'm not so sure I understand you.  I have removed paragraphs "not mention what is not", but still I'm afraid I don't properly get your point.

mfrasca: the point, later in the docs, where anonymous functions are mentioned along with named functions is in https://elixir-lang.org/getting-started/modules-and-functions.html#function-capturing

given the 'no parens' deprecation, and the still applying need to use `&` capture operator, explaining the current situation feels a bit awkward.  but I'm talking about a different spot in the docs.

josevalim: There have been long discussions on the topic (you can find those online) but the short answer is that `&` will still exist. The deprecation of `add` as a function call is only to avoid ambiguity with variables.

josevalim: There have been long discussions on the topic (you can find those online) but the short answer is that `&` will still exist. The deprecation of `add` as a function call is only to avoid ambiguity with variables.

josevalim:

PR #1267, opened 2019-06-05
comparing named/anonymous functions

mfrasca:
I'm not so sure of this edit, but it does make more sense to me this way.  I might be misunderstanding radically though.  

I've added a back-link to the place where anonymous functions had been introduced, and named functions mentioned.

I state (please correct if wrong) that there is no difference in the potential of anonymous vs. named functions.

I explain the capture operator `&` as a sort of function anonymiser.  too many words?  too few?  it does help my understanding, this way.

### DIFF
--- a/getting-started/modules-and-functions.markdown
+++ b/getting-started/modules-and-functions.markdown
@@ -89,7 +89,11 @@ The file will be compiled in memory and executed, printing "3" as the result. No
 
 ## Named functions
 
-Inside a module, we can define functions with `def/2` and private functions with `defp/2`. A function defined with `def/2` can be invoked from other modules while a private function can only be invoked locally.
+We briefly mentioned named functions when handling the [anonymous function basic type](/getting-started/basic-types.html#anonymous-functions). Here that we are discussing modules, it is the place for their named counterpart.
+
+Inside a module, we can define public named functions with `def/2` and private named functions with `defp/2`. A public named function, i.e.: one defined with `def/2`, can be invoked from other modules, while a private function can only be invoked locally.
+
+Apart from the syntax regarding their identifiers, there is no difference between named and anonymous functions, so all that we show here, applies anonymous functions as well.
 
 ```elixir
 defmodule Math do
@@ -159,7 +163,7 @@ iex> fun.(0)
 true
 ```
 
-Remember Elixir makes a distinction between anonymous functions and named functions, where the former must be invoked with a dot (`.`) between the variable name and parentheses. The capture operator bridges this gap by allowing named functions to be assigned to variables and passed as arguments in the same way we assign, invoke and pass anonymous functions.
+When introducing anonymous functions, we associated a variable name to an anonymous function definition, and we had to invoke the function with a dot (`.`) between the variable name and parentheses. This has nothing to do with the function being defined anonymous, and all to do with the function being accessed through a variable. The capture operator `&` allows named functions to be assigned to variables and to be passed as arguments in the same way we assign, invoke and pass anonymous functions.
 
 Local or imported functions, like `is_function/1`, can be captured without the module:
 


josevalim:
There are actually many differences between named and anonymous functions. Anonymous functions are closures. Named ones are not. Named functions can be serialized across nodes when captured, anonymous functions cannot.

So while some of these are advanced concerns which are not worth discussing at this point, I would not say they are the same.

In any case, thanks for the PR!

mfrasca:
but wait, before closing the PR … that's point 2 which we could omit/correct, but what about the back-link and the explanation around the `&` operator?

josevalim:
Sorry, I think my previous reply was not clear, as I meant to answer the second question too. I don't think & is a function anonymizer either, because the name is kept after all.

```elixir
iex(1)> &is_atom/1
&:erlang.is_atom/1
```

I am indifferent to the back link. If we want to add one, perhaps we should do it when we talk about the capture operator, instead of referring back to the chapter twice?

michalmuskala:
@josevalim you can send both captured named functions and anonymous functions across nodes - the limitation in the second case is, though, that the same module with the same version has to be present at both nodes so that you can call the transmitted anonymous functions.

mfrasca:
I don't know what would be the best place for the back link.  my impression was, as soon as we mention named functions, for we had referred to them when introducing anonymous functions.

it was my mistake in the comment to the commit, to call the `&` "anonymiser".  I do not use this term in the pushed text itself.  have a second look, then see it yourself I would say.  I'll go on with reading.

have you got some links to the discussion about the `&` operator?  once **naming** a function does not **invoke** it, I fail to see the reason for the `&` and the `.` (or it's because Elixir keeps variables and functions in two different tables?)

michalmuskala:
@mfrasca yes, elixir is a lisp-2 language which means that variables and functions are in different "namespaces". This means you can have local variables with names of local functions without any issues, but it also means you need different syntax when you want to call a function contained in a variable. For example consider:
```elixir
bar = &foo/0
bar.() #=> 1
bar() #=> 2

def foo(), do: 1
def bar(), do: 2
```

josevalim:
@michalmuskala right. Perhaps a better term is "safely serialized"?

@mfrasca unless linking back is important for understanding the section, I wouldn't do so to not affect the flow. I am also not a fan of calling named functions as "counterparts of anonymous functions".

Trying to associate the `.` with a variable, as implied in the last change in the diff is not fully precise either, as the dot is necessary in cases like this `(fn a, b -> a + b end).(1, 2)` and this `(&+/2).(1, 2)`, even though there are no variables.

mfrasca:
about the `.` … what is then, from a grammar point of view, on the left side of the `.`?  a variable matched to a function (anonymous or not), an anonymous function, and how do you call the form `(&(&1+&2))`?

michalmuskala:
An expression that evaluates to a function value.

josevalim:
Btw, before I forget, I want to make clear your PRs are very appreciated. :heart: Sometimes we do close them fast, but we often merge them fast too. :)

mfrasca:
I'll keep reading.  I'm not too happy with the current text in this spot.  I mean, it's a getting started, I see that, but I would not protect the reader from too much depth.

like we are mentioning both named/anonymous functions, but what they have in common (grammar and semantic) and what distinguishes them, it's not even suggested.  and yes, the closureness, and serializableness, not going to explain here, but links or hints might be well placed.

»elixir is a lisp-2 language« sounds worth mentioning, as well as that funny example showing the `.()` vs. `()` syntax.

mfrasca:
> An expression that evaluates to a function value.

and what is it before the `()`?

michalmuskala:
A name of a function - either local or remote (in a different module).

I don't think talking about "lisp 2" in an introductory material will be very useful - it's a term that most probably the majority of people won't be familiar - I think that by introducing such terms we only make the thing harder, not easier, to understand.

mfrasca:
I am seeing the difficulty, between being correct, and being understandable.  please help my understanding, and I think I will drop this point altogether afterwards.

We mentioned in [Anonymous functions](/getting-started/basic-types.html#anonymous-functions), that Elixir makes a distinction between anonymous functions and named functions. Apart from deeper details, one difference is that when we access a named function, we do so through a table — a "namespace" — which contains guaranteed only function definitions. When we evaluate an anonymous function, we have a "first class citizen", and we have to assert it is a function, before we can evaluate it. We do so with the intermediate dot (`.`) between the variable name (or function pointer) and the parentheses.

The capture operator bridges this gap by allowing named functions to be assigned to variables.  When we do so, we access named functions through "first class citizens", and we must handle it in the same way we assign, invoke and pass anonymous functions.

michalmuskala:
The distinction isn't as much between named and anonymous functions - it's between functions and function values. The `&` operator is how you convert a regular function to a function value, the `.(...)` operator is how you call a function value. You can also create function values (this time full closures) with the `fn` construct. 

The operator to call a function is `()` while the operator to call a function value is `.()`. Most languages compound those two operators into the same one, but it's not needed - you can keep them separate. Those languages that compound the two operators into one have later some heuristics to figure out which operation to execute. Similar how some languages have `+` for integer addition and `.+` for float addition and others compound those two operators into a single one and later have some logic to figure out which operation to actually execute.

mfrasca:
when I said "but why", I was not keeping in account your hint and funny example, where we have the same name `bar` in the two tables, or namespaces (or what is the correct term in erlang-speak?).  so there's no heuristics that holds.
you're saying two distinct operators, `()`, and `.()`.  ok, this helps understanding.  it's not first the dot and then the `()` operator.

mfrasca:
We mentioned in [Anonymous functions](/getting-started/basic-types.html#anonymous-functions), that Elixir makes a distinction between (anonymous) function values and (named) functions. Apart from deeper details, one difference is that when we use a named function, we are reaching its definition through a table — a "namespace" — which contains guaranteed only function definitions. The obvious thing to do with such an object is to evaluate it (automatic evaluation is now deprecated in Elixir). The operator to explicitly evaluate a function is `()`.

When we reach a function definition through a value, the operator to evaluate it is `.()`, that is: including the leading `.`.

The capture operator bridges this gap by extracting the function value from a (named) function, allowing it to be matched to a variable. When we use the matched variable, we are accessing a function value, so we must handle it in the same way we assign, invoke and pass anonymous functions.

mfrasca:
but actually, I'm just trying to understand the minimum necessary to evaluate whether or not to rewrite a django app in phoenix.  :scream: :palm_tree:

mfrasca:
```iex
iex> defmodule TemporaryContainer do
...>   def foo, do: "TC.foo"
...>   def bar, do: "TC.bar"
...> end
{:module, TemporaryContainer,
 <<70, 79, 82, 49, 0, 0, 4, 76, 66, 69, 65, 77, 65, 116, 85, 56,
   0, 0, 0, 140, 0, 0, 0, 14, 25, 69, 108, 105, 120, 105, 114, 46,
   84, 101, 109, 112, 111, 114, 97, 114, 121, 67, 111, 110, 116,
   97, 105, ...>>, {:bar, 0}}
iex> import TemporaryContainer
TemporaryContainer
iex> bar = &foo/0
&TemporaryContainer.foo/0

# this is deprecated syntax, evaluating foo
iex> foo
"TC.foo"
# this is the proper way to evaluate foo
iex> foo()
"TC.foo"

# this evaluates the function value associated to bar
iex> bar.()
"TC.foo"
# this evaluates the function defined as bar
iex> bar()
"TC.bar"
# returning the value takes precedence above evaluating the function
iex> bar
&TemporaryContainer.foo/0
```

PR #1270, opened 2019-06-06
for readers wondering what's the use of atoms

mfrasca:


josevalim:
Thanks. Although isn’t this already covered by the first example comparing
:hello and :world?
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

mfrasca:
ah, sorry, yes, and also no.  I mean, it's clear, but not explicit.  I'll review.

mfrasca:
@josevalim , question, related but not strictly.  where to discuss the shape of the 'getting started', more than the tiny details.  like, should it point earlier, or more prominently, to how to structure a program?  for example giving a reasonable example of a tiny app with more than two processes.  (maybe a RPN calculator, with logging, similar to the emacs `calc`).  right, maybe I missed it.

josevalim:
Here is the place. :)

I would just say that’s not the focus of the Getting Started guide. I
understand the need for more detail, more examples, but that’s not what the
guide is about. It is not that your approach is wrong and this is right or
vive-versa they are just different approaches. The guide is about a quick
discussion of the main language concepts, to put you on a path for whatever
you want to do next.

The Mix & OTP guide does cover building an actual application though, with
a lot of detail.


-- 


mfrasca:
> The Mix & OTP guide does cover building an actual application though, with a lot of detail.

had seen that … my impression was indeed lots of detail, but no clear "political/architectural" explanation.  I'll have another look at it.  I mean: if the strongest point here is concurrency, let's show how to structure a simple program where concurrency is in the spotlight.  but sometimes I don't see things even when in full view.

josevalim:
I like the description of enumeration, so i pushed more examples here: 15267d78f07636a437d098ca282a8cd760fe5ced

PR #1271, opened 2019-06-08
reviewing getting_started/GenServer

mfrasca:
I was a bit overwhelmed when the `getting started` guide reached the use of `Agent` and even more so when it was the time for `GenServer`.  I added an example of plain straight use of Agent, which helped me make sense of writing and testing the KV.Bucket API.  I later saw this is precisely the approach taken in the https://hexdocs.pm/elixir/GenServer.html linked page.  I can follow it better that way.

but now what, I mean, do we want the reader to jump to the other docs, read and understand, but with the chance they get lost in far deeper details?  or do we want to copy/adapt part of that text into this getting-started guide?  I still didn't reach the place where it handles supervisors.

### DIFF
--- a/getting-started/mix-otp/agent.markdown
+++ b/getting-started/mix-otp/agent.markdown
@@ -71,7 +71,9 @@ iex> Agent.get(agent, fn content -> content end)
 iex>
 ```
 
-It is completely up to us, to use `Agent` in a sensible way, and the proper way to go is by writing modules that enforce a behaviour, by exposing a well defined API. Let's do precisely that, let's implement our `KV.Bucket` using agents and let's start by first writing some tests, to define the API exposed by our module. Create a file at `test/kv/bucket_test.exs` (remember the `.exs` extension) with the following:
+What does the above mess show? The sensible use of agents is completely in our hands, and the proper way to go is by writing modules that implement a behaviour, by exposing a well defined API. Let's do precisely that, let's implement our `KV.Bucket` using `Agent` and let's start by first writing some tests, to define the API exposed by our module.
+
+Create a file at `test/kv/bucket_test.exs` (remember the `.exs` extension) with the following:
 
 ```elixir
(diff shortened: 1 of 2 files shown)


mfrasca:
there's one point which (I'm sure) I do not follow: 
```
    KV.Registry.create(registry, "shopping")
    assert {:ok, bucket} = KV.Registry.lookup(registry, "shopping")
```

and create is implemented as a `cast` and lookup as a `call` and we said that casts are asynchronous.  how can we make sure the cast was executed before the subsequent call?

josevalim:
If process A sends process B two messages, the first will always arrive
before the second. The cast and call is not about the sending but rather
about the receiving. In particular, cast receives no reply, but call blocks
until a reply is received.
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

mfrasca:
I think I got that.  and I guess that computers won't stay idle or waiting for nothing as mine sometime does (WAI:94%), but I send a cast, receive no answer, send a call, wait for an answer, how do I know, before I send the call, that the cast was executed?

josevalim:
Each process has a message inbox. Agent, GenServer, etc all process the
messages in order.
-- 


mfrasca:
an other point (apart from the above "what about `@impl`?"), is that I'm not sure I so quickly see the point of doing all this with `GenServer` instead of with a flat `Agent`.  it's not yet visible in this page or is it?

josevalim:
Thanks @mfrasca! Those changes are superb. I did one final pass over your edits, to make sure they all follow the same "tone". Not that one tone is better than the other (or maybe they are, but I wouldn't know which :D), but just to make sure it reads like it was written by one person.

josevalim:

mfrasca:
hi, thanks for the edits, I had to smile wide when I saw how you so neatly presented "the above mess".

PR #1272, opened 2019-06-08
stress the difference use/(alias, import, require)

mfrasca:
understood this when reached GenServer.

### DIFF
--- a/getting-started/alias-require-and-import.markdown
+++ b/getting-started/alias-require-and-import.markdown
@@ -132,7 +132,7 @@ Note that `import`ing a module automatically `require`s it.
 
 ## use
 
-The `use` macro is frequently used as an extension point. This means that, when you use a module `FooBar`, you allow that module to inject *any* code in the current module, such as importing itself or other modules, defining new functions, setting a module state, etc.
+The `use` macro is frequently used as an extension point. This means that, when you `use` a module `FooBar`, you allow that module to inject *any* code in the current module, such as importing itself or other modules, defining new functions, setting a module state, etc. 
 
 For example, in order to write tests using the ExUnit framework, a developer should use the `ExUnit.Case` module:
 
@@ -146,7 +146,11 @@ defmodule AssertionTest do
 end
 ```
 
-Behind the scenes, `use` requires the given module and then calls the `__using__/1` callback on it allowing the module to inject some code into the current context. Generally speaking, the following module:
+Behind the scenes, `use` requires the given module and then calls the `__using__/1` callback on it allowing the module to inject some code into the current context. Some modules (e.g., the above `ExUnit.Case`, but also `Supervisor` and `GenServer`) use this mechanism to populate your module with some basic behaviour, which your module is intended to override or complete. 
+
+Whether you can `use` a module, it depends on that module defining a `__using__/1` function. To `use` a module only makes sense within an other module.
+
+Generally speaking, the following module:
 
 ```elixir
 defmodule Example do
@@ -163,7 +167,7 @@ defmodule Example do
 end
 ```
 
-Since `use` allows any code to run, we can't really know the side-effects of using a module without reading its documentation. For this reason, `import` and `alias` are often preferred, as their semantics are defined by the language.
+Since `use` allows any code to run, we can't really know the side-effects of using a module without reading its documentation. For `import` and `alias`, their semantics are defined by the language. `use` serves quite a different purpose.
 
 ## Understanding Aliases
 
--- a/getting-started/alias-require-and-import.markdown
+++ b/getting-started/alias-require-and-import.markdown
@@ -132,7 +132,7 @@ Note that `import`ing a module automatically `require`s it.
 
 ## use
 
-The `use` macro is frequently used as an extension point. This means that, when you `use` a module `FooBar`, you allow that module to inject *any* code in the current module, such as importing itself or other modules, defining new functions, setting a module state, etc. 
+The `use` macro is frequently used as an extension point. This means that, when you `use` a module `FooBar`, you allow that module to inject *any* code in the current module, such as importing itself or other modules, defining new functions, setting a module state, etc.
 
 For example, in order to write tests using the ExUnit framework, a developer should use the `ExUnit.Case` module:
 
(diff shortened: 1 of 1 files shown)


getting-started/alias-require-and-import.markdown
```diff
@@ -132,7 +132,7 @@ Note that `import`ing a module automatically `require`s it.
 
 ## use
 
-The `use` macro is frequently used as an extension point. This means that, when you use a module `FooBar`, you allow that module to inject *any* code in the current module, such as importing itself or other modules, defining new functions, setting a module state, etc.
+The `use` macro is frequently used as an extension point. This means that, when you `use` a module `FooBar`, you allow that module to inject *any* code in the current module, such as importing itself or other modules, defining new functions, setting a module state, etc.
```
whatyouhide:
```suggestion
The `use` macro is frequently used as an extension point. This means that, when you `use` a module `FooBar`, you allow that module to inject *any* code in the current module, such as importing itself or other modules, defining new functions, setting a module state, etc.
```

getting-started/alias-require-and-import.markdown
```diff
@@ -146,7 +146,11 @@ defmodule AssertionTest do
 end
 ```
 
-Behind the scenes, `use` requires the given module and then calls the `__using__/1` callback on it allowing the module to inject some code into the current context. Generally speaking, the following module:
+Behind the scenes, `use` requires the given module and then calls the `__using__/1` callback on it allowing the module to inject some code into the current context. Some modules (e.g., the above `ExUnit.Case`, but also `Supervisor` and `GenServer`) use this mechanism to populate your module with some basic behaviour, which your module is intended to override or complete.
```
whatyouhide:
```suggestion
Behind the scenes, `use` requires the given module and then calls the `__using__/1` callback on it allowing the module to inject some code into the current context. Some modules (for example, the above `ExUnit.Case`, but also `Supervisor` and `GenServer`) use this mechanism to populate your module with some basic behaviour, which your module is intended to override or complete. 
```

getting-started/alias-require-and-import.markdown
```diff
@@ -146,7 +146,11 @@ defmodule AssertionTest do
 end
 ```
 
-Behind the scenes, `use` requires the given module and then calls the `__using__/1` callback on it allowing the module to inject some code into the current context. Generally speaking, the following module:
+Behind the scenes, `use` requires the given module and then calls the `__using__/1` callback on it allowing the module to inject some code into the current context. Some modules (e.g., the above `ExUnit.Case`, but also `Supervisor` and `GenServer`) use this mechanism to populate your module with some basic behaviour, which your module is intended to override or complete. 
+
+Whether you can `use` a module, it depends on that module defining a `__using__/1` function. To `use` a module only makes sense within an other module.
```
whatyouhide: > To `use` a module only makes sense within an other module.

I'm not sure I understand what you mean here.

mfrasca: you can execute any of the `alias`, `require`, `import` macros from the `iex` prompt.
but `use` only makes sense to inject code into a module.

mfrasca: this part was the "busillis" which I had missed, and forms the core of my edit, I think.

whatyouhide: You can use `use MyMod` outside of a module because it's exactly like calling `MyMod.__using__([])` which is just a macro, and you can call macros from IEx (like `Logger.info/1` to name one). It's just that `GenServer`, `Supervisor`, and many other modules do things inside `__using__` that can only be done inside of modules, such as setting `@module_attributes`. Does that make sense?

mfrasca: yes. … … not sure whether to remove that short sentence, or to make it sound like:
»several modules do things inside `__using__` that can only be done inside of modules, so that `use` may cause an error if executed outside of a module.«
which will possibly only confuse the reader.  however correct.

mfrasca: changed the text, have a look if you have the time.

josevalim:
All looks good except the last paragraph which was harder to read, so i reverted that one and kept the rest, thanks!

josevalim:

mfrasca:
thanks for reviewing and merging, but let me explain the reason behind changing the last paragraph.
it's not a matter of "preferring", it's a matter of knowing what it does.
if I may compare Elixir with Python, this `use` is the closest thing to subclassing, and it's not that I "prefer" subclassing when I write my `class LocalTest(unittest.TestCase)`, it's really the only proper thing to do there.
on the other hand obviously, I would not subclass something I know nothing about.
same here, with `use`.  it's **the** tool to define a test module, or my `Agent`, etc.
you got the point, I'm sure.  I'll stop here.

josevalim:
Well, let me rephrase my feedback. My understanding of the last sentence was not adding anything new to what was being said today. So I am not quite sure what you mean yet. Could you please give it another try at explaning it to me? Thanks.

And to be clear, use is quite different from subclassing. I mean, it could be used to emulate subclassing but even for things like `GenServer`, we are not overriding a default implementation, we are rather "defining callbacks".

mfrasca:
let's stay on the `GenServer` example.  that module describes a behaviour, and places slots which we can override.  sorry, again the "override" which you replaced by "define", I keep using OO terminology, since this is how I understand it.  anyhow, we want a concrete server which is less generic than GenServer, so we fill in the blanks where concrete behaviour can be defined.  according to me, to people coming from OO, `use` starts to make sense in this perspective.  I do understand it's not the same, but if you tell me, »look, learn defining `__using__` and taking advantage of `use`, because they will enable to write modules serving as Abstract Base Classes, and `use` will serve you as a tool for subclassing«, it will turn on a light.
I can imagine that my mostly OO programming limits me in seeing other possible uses of this `use`, but within the limits of my experience, this helps making sense of `use`.

mfrasca:
p.s.: "defining callbacks" reminds me of implementing abstract methods when subclassing an abstract base class.  I understand that the GenServer callbacks do not throw "not implemented" exceptions, so it's not the same, but it does remind me of that.

josevalim:
What you said makes sense! But in the paragraph before the last one, we do talk about `__using__` and how `use` invokes it, so doesn't it contain all information necessary?

> p.s.: "defining callbacks" reminds me of implementing abstract methods when subclassing an abstract base class. I understand that the GenServer callbacks do not throw "not implemented" exceptions, so it's not the same, but it does remind me of that.

That's the closest, yeah!  We don't raise but we do warn (for example, if you don't define `init/1`). The other callbacks are optional though and raise at runtime if you don't implement them.

mfrasca:
> […] in the paragraph before the last one, we do talk about `__using__` and how `use` invokes it, so doesn't it contain all information necessary?
> 

possibly … although I would say: ›not explicit enough‹, and I would wonder: what's this statement about 'preferring' telling me? it's not a matter of preferences, is it?

PR #1275, opened 2019-06-10
removing link to proprietary cheat sheet

mfrasca:
I don't think this cheat sheet makes things so much clearer:
- it has hardly any structure (like it would be if all client->return-value-from-client->callback->return-value-from-callback would all be organized as in the clearly structured initialization block), 
- graphic elements drawn on top of the text make it less readable, 
- I think it has flaws (not mentioning the return value from `handle_call`, explaining `*`, why return value tuples have different sizes) 

**and**
- it has one major problem: the license does not allow correcting it.

### DIFF
--- a/getting-started/mix-otp/genserver.markdown
+++ b/getting-started/mix-otp/genserver.markdown
@@ -322,8 +322,6 @@ So far we have used three callbacks: `handle_call/3`, `handle_cast/2` and `handl
 
 Since any message, including the ones sent via `send/2`, go to `handle_info/2`, there is a chance unexpected messages will arrive to the server. Therefore, if we don't define the catch-all clause, those messages could cause our registry to crash, because no clause would match. We don't need to worry about such cases for `handle_call/3` and `handle_cast/2` though. Calls and casts are only done via the `GenServer` API, so an unknown message is quite likely a developer mistake.
 
-To help developers remember the differences between call, cast and info, the supported return values and more, [Benjamin Tan Wei Hao](http://benjamintan.io) has created an excellent [GenServer cheat sheet](https://raw.githubusercontent.com/benjamintanweihao/elixir-cheatsheets/master/GenServer_CheatSheet.pdf).
-
 ## Monitors or links?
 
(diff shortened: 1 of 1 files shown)


josevalim:
It is a cheat sheet, it is meant to be succinct and not provide much information. In any case, the cheat sheet does not have to make sense for everyone, but i believe in many cases it likely helps, so I would keep it around. :) Thanks!

mfrasca:
as said: the main problem with this cheat sheet is not the content, it's the **license**.
I'll see if I manage to produce something about both `Agent` and `GenServer`, and it will be CC:BY-SA.
so we can talk about it constructively.

whatyouhide:
I'm fairly sure @benjamintanweihao is pretty open to letting us use this :)

mfrasca:
> I'm fairly sure @benjamintanweihao is pretty open to letting us use this :)

right, I did not ask him to change the license.  I'll do so before I start something from scratch.  as it stands, no derivative work is allowed, no commercial use either.

mfrasca:
https://github.com/benjamintanweihao/elixir-cheatsheets/issues/2

mfrasca:
oops, too quick in deleting the branch, how do I add to this now …

benjamintanweihao:
@whatyouhide @mfrasca Ohai! I'm totally OK with changing the license and providing the source. Please give me a day or two. ❤️

mfrasca:
hi Author @benjamintanweihao ! I've been studying your cheat sheet and I thought I would use it for teaching myself some latex+tikz (it was on my todo list).  result is sort of the same as what you had shared, but more numerical and somewhat lighter (pdf is ~50k, instead of 350k).  if you want to have a look?  #1276.  I'm fine with whatever works, and you bet I prepare my coffee from emacs.

benjamintanweihao:
I'm happy to look at the final PDF. I prepared the cheatsheet in something less sophisticated.: MS Powerpoint.

mfrasca:
that's nice.  I put both the tex and the pdf in the #1276, have a look, I have several questions.  keep in mind 2 weeks ago I had not yet heard of elixir.  and I had never given much thought to erlang, mostly because of the weird syntax.  I had studied miranda/haskell but never used it in real life.

PR #1276, opened 2019-06-10
rework cheat sheet

mfrasca:
just in case there won't be an answer from the author, I've been making this one.
and although I'm not sure it is any better than the now linked pdf, I am positive I can edit it right now if you have fixes.


mfrasca:
I do not understand, and I did not look up, the role of that third coordinate in the `return_value` tuple, that `5_000`, or `:hibernate`.

I have left three slots open, which aren't handled in the getting-started, and were not in the original cheat sheet.
  * GenServer.code_change/3 (function)
  * GenServer.format_status/2 (function)
  * GenServer.handle_continue/2 (function)

I'm not so sure about repeating stuff, instead of defining it in one place, and then refer to it.  this is needed for both `reason` and `return_value`, and I'm also not so sure about the practice of giving one concrete value and leaving to the reader the interpretation "I'm writing here `{:async_op, args}`, but you are free to put whatever you like here, as long as you catch/match it in your callbacks".  I tried to remove this practice, but it's still there in at least two spots.

mfrasca:
I replaced 'return_value' with 'result'.  makes a shorter title in the box with vertical text.

josevalim:
Hi @mfrasca! This looks great! A couple suggestions for now:

1. "Immediate" is not clear. Maybe we should call it "Return"? Or "Result"? Or "Client Return"?

2. The stop reason is not only `:normal`, `:shutdown`, and `{:shutdown, term}`. Those are the reasons with special meaning regarding termination but any value is supported.

3. The `:stop` for call is `{:stop, reason, reply, state}`. It would be cool if we do an arrow from the `reply` in the result to the reply in the "Return" section

3. The `:stop` for cast and info is `{:stop, reason, state}`

mfrasca:
immediate, after all, it's just a name.  I'm using it without defining it, which obviously is not the correct thing to do.  I could change it to "client return", but would that mean we don't need to define that either?  graphically, 'immediate' is shorter.

about the other points, we are describing what is defined in `GenServer`, the programmer being free to define their own meanings all over the place, isn't it?  I think this could go in a the missing explanation.

I do not understand: `{:shutdown, term}`, nor the role of `state` in a `:stop`.

josevalim:
Immediate is not used anywhere in Elixir, so even if we admit the word
choice doesn’t matter much, we need to pick at least a familiar one. It is
fine to call it “Returns”. Maybe we can have a visual indication that it is
the client return somehow? Start_link also has other return types, such as
:ignore or {:error, ...}.

If you stop a server after it is initialized, terminate is called with the
state, that’s why all callbacks except init expect a state in stop.

If you want to keep a summary of the stop reasons, I would keep this:

* normal - doesn’t log, doesn’t break links
* shutdown, {shutdown, term} - doesn’t log, breaks links
* anything else - logs, breaks links

“term” just means any value. We could also use “any”, “value”, etc.
-- 


mfrasca:
> Maybe we can have a visual indication that it is the client return somehow?

yes, that was my thought already.  I put a first page with init, terminate, handle_cast (in these cases the corresponding `start`, `stop`, and `cast` functions return immediately).  and a second page where the `call` does not return until there is a reply from the callback.  and I moved the box next to the one explaining how to match the result from the callback.

I like calling it `returns`.

mfrasca:
right, have a look.  and we still miss:

- [ ] what is that third coordinate in `result`? (5000, :hibernate)?
- [ ] what about code_change, format_status, handle_continue?
- [ ] do we want some introduction/explanation/header?

getting-started/mix-otp/genserver.markdown
```diff
@@ -322,7 +322,7 @@ So far we have used three callbacks: `handle_call/3`, `handle_cast/2` and `handl
 
 Since any message, including the ones sent via `send/2`, go to `handle_info/2`, there is a chance unexpected messages will arrive to the server. Therefore, if we don't define the catch-all clause, those messages could cause our registry to crash, because no clause would match. We don't need to worry about such cases for `handle_call/3` and `handle_cast/2` though. Calls and casts are only done via the `GenServer` API, so an unknown message is quite likely a developer mistake.
 
-To help developers remember the differences between call, cast and info, the supported return values and more, [Benjamin Tan Wei Hao](http://benjamintan.io) has created an excellent [GenServer cheat sheet](https://raw.githubusercontent.com/benjamintanweihao/elixir-cheatsheets/master/GenServer_CheatSheet.pdf).
+To help developers remember the differences between call, cast and info, the supported return values and more, we have a tiny [GenServer cheat sheet](https://raw.githubusercontent.com/elixir-lang/elixir-lang.github.com/master/cheat-sheet.pdf), and it comes with `tex`/`tikz` source.
```
josevalim: This will be hosted on the website, so we should do the website link. :) Maybe we should create a cheatsheets directory at the project root and have this called and linked as `cheatsheets/gen-server.pdf`.

josevalim:
> what is that third coordinate in result? (5000, :hibernate)?

Each option is disconnected from the other. 5000 is a timeout (you will get a message if nothing happens within the given time in miliseconds) and the :hibernate tells the process to hibernate (compact memory and go sleep). I would use `timeout` instead of 5000.

>  what about code_change, format_status, handle_continue?

`code_change` and `format_status` is rarely used. `handle_continue` is relatively new. In the same place you pass `timeout` or `:hibernate`, you can pass `{:continue, value}`, and `handle_continue(value, state)` will be invoked

>  do we want some introduction/explanation/header?

I don't think so. My suggestion is to link to `https://hexdocs.pm/elixir/GenServer.html`.

getting-started/mix-otp/cheat-sheet.tex
```diff
@@ -0,0 +1,350 @@
...
+:ok
+\end{verbatim}
+};
+  \path (result.north west)+(-1em,0) node [left-title,anchor=north] {\rotatebox{90}{returns}};
+
+  \path (client.south west)+(0,-1em) node (callback) [main-block,anchor=north west] {
+    \vspace*{-\baselineskip}
+\begin{verbatim}
+def handle_cast(match_this, state) do
+  # compute new state or fail to do so
```
josevalim: The "fail to do so" is ambiguous. Does it mean an error? If so, I don't think we should say something can fail, because anything can fail at any time. We could say:

```suggestion
  # Returns noreply or stop with new state
```

Thoughts?

mfrasca: it was meant to be ambiguous.  I would not mention 'returns whatever', because that's the following line that specifies the value of the evaluation of `handle_cast`.  the meaning of the `result` is explained in the following box, the one with the `^result =` title.

josevalim:
So maybe we should remove the comment them, because it doesn’t add anything
new? We are not even required to compute a new state.
-- 


mfrasca:
I would not remove it.  it suggests "do whatever you need to", and it ends with `result`, which you need to provide.  compare to `terminate`.

josevalim:
So we need to use something generic like that instead of something
ambiguous. “# Do some work and compute result” would work.
-- 


mfrasca:
I'll commit it again, but I'm a bit short of vertical space, and I think we are repeating the same info all over the place, at least for the nil/timeout/:hibernate (and now I should also add `{:continue, value}`

mfrasca:
(you probably want to review your signature, for it doesn't work properly, the embedded link.)

getting-started/mix-otp/cheat-sheet.tex
```diff
@@ -0,0 +1,385 @@
...
+:ignore!
+\end{verbatim}
+};
+\path (result.north west)+(-1em,0) node [left-title,anchor=north,fill=green!25] {\rotatebox{90}{\^{}result = }};
+
+  \draw[color=blue!10!yellow!5,draw=blue!25!yellow,dashed]
+        (result.north west)+(0,-38pt) -- ($ (result.north east)+(0,-38pt) $);
+
+\path (result.south east)+(3em,0) node (reason) [immediate,anchor=south west,fill=yellow!30] {
+  This applies wherever \verb|result| matches \verb|{:stop, reason}|, and to \verb|reason| argument in \verb|stop| message.
```
josevalim: Can we say something like:

> Whenever there is a `:stop` tuple, the `reason` may be:

It could be nice to add the "log" / "break links" i commented earlier. If there is no space to write all of that, I would write this instead:

> If stopping due to expected scenarios, use `:shutdown`, `{:shutdown, any}`. If stopping due to an error, use any other value except `:normal`.

My main concern is that by listing `:normal`, people will most likely use, and you rarely want to use `:normal`.

getting-started/mix-otp/cheat-sheet.tex
```diff
@@ -0,0 +1,385 @@
...
+{:noreply, state}
+{:noreply, state, then_what}
+
+{:stop, reason, state}
+\end{verbatim}
+};
+  \path (result.north west)+(-1em,0) node [left-title,anchor=north,fill=green!25] {\rotatebox{90}{\^{}result =}};
+
+  \path (result.south east)+(3em,0) node (then-what) [immediate,anchor=south west,fill=yellow!30] {
+    This applies wherever \verb|result| matches \verb|{_, state, then_what}|
```
josevalim: `then_what` also appears in other places. that do not match `{_, state, then_what}`. Therefore I would say something like:

> `then_what` in this cheat-sheet means one of:

getting-started/mix-otp/cheat-sheet.tex
```diff
@@ -0,0 +1,385 @@
...
+
+\begin{verbatim}
+milliseconds
```
josevalim:
```suggestion
timeout_in_milliseconds
```

getting-started/mix-otp/genserver.markdown
josevalim: Btw, one of the reason to keep this in a separate directory is because we can add a `README` to said directory with instructions on how to build the cheat-sheets. :)

getting-started/mix-otp/cheat-sheet.tex
```diff
@@ -0,0 +1,385 @@
...
+\end{verbatim}
+};
+
+  \path (result.north west)+(-1em,0) node [left-title,anchor=north,fill=green!25] {\rotatebox{90}{\^{}result =}};
+
+  \draw[color=blue!10!yellow!5,draw=blue!25!yellow,dashed]
+        (result.north west)+(0,-38pt) -- ($ (result.north east)+(0,-38pt) $);
+
+  \path (result.south east)+(3em,0) node (then-what) [immediate,anchor=south west,fill=black!5] {
+Functions where \verb|@impl true| applies.
```
josevalim: I see you moved this to the side, so you can keep everything in the same page. I would try to keep it still at the bottom of the page. Maybe we can break it in a bullet list like this:

> * `@impl true` can be used before each definition to make sure they match the proper GenServer callback.
> * Other callbacks are: `code_change` and `format_status`
> * Last version: ...
> * Reference: ...
> * LICENSE

josevalim: Something else you could do to gain some space is to replace `handle_call` with `handle_cast`, since `handle_call` takes more vertical space and there is some vertical space left in the first page?

getting-started/mix-otp/cheat-sheet.tex
```diff
@@ -0,0 +1,385 @@
...
+  \path (callback.north west)+(-1em,0) node [left-title,anchor=north] {\rotatebox{90}{callback}};
+
+  \path (callback.south west)+(0,-1em) node (result) [main-block,anchor=north west] {
+    \vspace*{-\baselineskip}
+\begin{verbatim}
+{:ok, state}
+{:ok, state, then_what}
+
+{:stop, reason}
+:ignore!
```
josevalim:
```suggestion
:ignore
```

getting-started/mix-otp/cheat-sheet.tex
```diff
@@ -0,0 +1,385 @@
...
+{:shutdown, any}
+any
+\end{verbatim}
+};
+\path (reason.north west)+(-1em,0) node [left-title,anchor=north,fill=green!25] {\rotatebox{90}{\^{}reason = }};
+
+\end{tikzpicture}
+\end{parchment}
+
+\begin{parchment}[termination: .stop $\rightarrow$ \bf\texttt{terminate/2}]
```
josevalim: I would also add a note such as:

> terminate/2 is also called when :stop is returned and in case of errors when `Process.flag(:trap_exit)` is true.

This note can be in a box to the side. WDYT?

getting-started/mix-otp/cheat-sheet.tex
mfrasca: > It could be nice to add the "log" / "break links" i commented earlier. If there is no space to write all of that, I would write this instead:

I'll have a look.  now I'm using the `verbatim` environment, which does not allow me do what you suggest.  I will see what other possibilities we have.

getting-started/mix-otp/cheat-sheet.tex
mfrasca: aha, right, I had missed the `{:reply, reply, state, then_what}` callback result from handle_call.

then I would remove the whole text, because the sentence "`then_what` in this cheat-sheet means one of:" is contained, at least that's my intention, in the 90° title `^then_what =`.

josevalim:
Sounds good. Note that init can also return {:ok, state, then_what}, so we
should probably move it all the way up!
-- 


getting-started/mix-otp/cheat-sheet.tex
mfrasca: the three in the first page share the "immediate return".

I'll see what I can do anyway.

mfrasca:
> Note that init can also return {:ok, state, then_what}, so we should probably move it all the way up!

"all the way up", within the limitation of rectangular shapes I am using, that's not feasible, because "all the way up" is where I put the `reason` box.
notice that the `returns` box is vertically aligned with the `client` box.
the **this applies wherever …** boxes are in a slightly different colour which in my intention would help the user notice them and not associate them to the rectangular shape where they happen to be placed.

mfrasca:
ops.  hit 'resolve' instead of 'accept'.  I had already corrected manually.

getting-started/mix-otp/cheat-sheet.tex
```diff
@@ -0,0 +1,385 @@
...
+  GenServer.start_link(__MODULE__, match_this, opts)
+end
+\end{verbatim}
+    };
+    \path (client.north west)+(-1em,0) node [left-title,anchor=north] {\rotatebox{90}{client}};
+
+    \path (client.north east)+(3em,0) node (immediate) [immediate,anchor=north west] {
+      \vspace*{-\baselineskip}
+\begin{verbatim}
+{:ok, pid}
```
josevalim:
```suggestion
{:ok, pid}
:ignore
{:error, term}
```

getting-started/mix-otp/cheat-sheet.tex
```diff
@@ -0,0 +1,403 @@
...
+
+% Define a few styles and constants
+\tikzstyle{left-title}=[draw, fill=blue!20, minimum width=1em, minimum height=5em,anchor=north east]
+\tikzstyle{top-title}=[draw, fill=blue!20, minimum width=5em, minimum height=1em]
+\tikzstyle{main-block} = [left-title, text width=30em, inner sep=8pt, fill=red!15!yellow!40, minimum height=5em, rounded corners]
+\tikzstyle{immediate} = [left-title, text width=17em, inner sep=8pt, fill=red!15!yellow!40, minimum height=5em, rounded corners]
+\tikzstyle{explain} = [left-title, text width=17em, inner sep=8pt, fill=black!5, minimum height=5em, rounded corners,anchor=south west]
+
+\begin{comment}
+\begin{parchment}[commentarii de bello gallico]
```
josevalim: I believe this one can be removed. :)

getting-started/mix-otp/cheat-sheet.tex
```diff
@@ -0,0 +1,403 @@
...
+\end{parchment}
+\end{comment}
+\section*{GenServer - a cheat sheet}
+last version: \verb|https://elixir-lang.org/getting-started/mix-otp/cheat-sheet.pdf|\\
```
josevalim: Let's move this information to the footnote?

josevalim:
I have dropped some comments. I am just not happy with `reason`. You did a perfect job but I don't think we stroke the proper balance on how much information to show. Maybe we should just say "^reason = user defined"? A proper explanation of reasons could be done in an eventual supervisor cheat sheet?

getting-started/mix-otp/cheat-sheet.tex
mfrasca: I've added a footnotes block, and I think there's space for some extra explanation, like what's those `^field=` title, and why some blocks are in a colour rather than an other.

mfrasca:
I guess all that the user may define qualifies as user defined, but the framework, i.e. the GenServer module (in my eyes, something approaching a base class, but not quite), it does define patterns and behaves accordingly, so that we better follow the GenServer defined patterns, rather than feel free as in "user defined".

josevalim:
@mfrasca one of the things worth saying though is that the reasons (`normal, shutdown, {shutdown, _}`) are not defined by GenServer, but mostly Erlang (normal) and the `Supervisor` module.

In any case, I think this is good to go IMO. We need only two remaining changes:

1. Move the cheatsheet to the `cheatsheet ` directory at the project root
2. Include a README file with dependencies and exact steps to build the cheatsheet :)

mfrasca:
I don't know if the README would be enough.  LaTeX has always been on all of my computers, and I tend to forget how to install it, I just use it.  and to be true, I don't even invoke it from the terminal, I generally have a two-panes emacs session where I'm editing the tex and whenever I save it, emacs takes care of producing the pdf and reload it in the other pane.

mfrasca:
> the reasons (`normal, shutdown, {shutdown, _}`) are not defined by GenServer, but mostly Erlang (normal) and the `Supervisor` module

sounds like we can put this in a footnote?

josevalim:
I will try reproducing these steps and building it locally. Regarding the
exit reasons, maybe we should just link it here:
https://hexdocs.pm/elixir/Supervisor.html#module-exit-reasons-and-restarts
-- 


josevalim:
This is amazing, thank you! :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

mfrasca:
> This is amazing, thank you! heart green_heart blue_heart yellow_heart purple_heart

well, thank you, but I'm really scared in front of the other cheat sheet (Supervisor).  I think I'll go back to the python/django world for a while.

PR #1277, opened 2019-06-11
introduce supervisors at the top of the supervisors chapter

mfrasca:
I was reorganizing the text at the end of the chapter, answering (proposing an answer) the the open question "if this is a bad idea, what would be the proper solution?".  and I realized I was not only closing the chapter, but also introducing supervisors, which I think is precisely the introduction I had missed to that chapter.

this is part of what I was writing, mixed with stuff I intend to remove.

This however has a major drawback: when our registry is notified about any unexpected crash, we have to program the whole reaction to the new situation, a more prevalent practice in We typically avoid creating new processes directly, instead, we delegate this responsibility to supervisors. 

As we'll see in the next chapter, supervisors rely on links and that explains why link-based APIs (`spawn_link`, `start_link`, etc) are so prevalent in Elixir and <abbr title="Open Telecom Platform">OTP</abbr>.

### DIFF
--- a/getting-started/mix-otp/genserver.markdown
+++ b/getting-started/mix-otp/genserver.markdown
@@ -337,4 +337,4 @@ Returning to our `handle_cast/2` implementation, you can see the registry is bot
 ref = Process.monitor(pid)
 ```
 
-This is a bad idea, as we don't want the registry to crash when a bucket crashes! We typically avoid creating new processes directly, instead, we delegate this responsibility to supervisors. As we'll see in the next chapter, supervisors rely on links and that explains why link-based APIs (`spawn_link`, `start_link`, etc) are so prevalent in Elixir and <abbr title="Open Telecom Platform">OTP</abbr>.
+This is a bad idea, as we don't want the registry to crash when a bucket crashes. A reasonable fix here, once we have settled for our `KV.Registry` to monitor its buckets via references, is replace `KV.Bucket.start_link` for a plain `start`. A more radical solution is explored in the next chapter, where we introduce "supervisors".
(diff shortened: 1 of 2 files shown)


josevalim:
Perfect. Your description helped me put things in motion, please let me know what you think!

getting-started/mix-otp/genserver.markdown
```diff
@@ -337,4 +337,4 @@ Returning to our `handle_cast/2` implementation, you can see the registry is bot
 ref = Process.monitor(pid)
 ```
 
-This is a bad idea, as we don't want the registry to crash when a bucket crashes! We typically avoid creating new processes directly, instead, we delegate this responsibility to supervisors. As we'll see in the next chapter, supervisors rely on links and that explains why link-based APIs (`spawn_link`, `start_link`, etc) are so prevalent in Elixir and <abbr title="Open Telecom Platform">OTP</abbr>.
+This is a bad idea, as we don't want the registry to crash when a bucket crashes. A reasonable fix here, once we have settled for our `KV.Registry` to monitor its buckets via references, is replace `KV.Bucket.start_link` for a plain `start`. A more radical solution is explored in the next chapter, where we introduce "supervisors".
```
josevalim: Using `start` is bad advice too. :) My suggestion:

```suggestion
This is a bad idea, as we don't want the registry to crash when a bucket crashes. The proper fix is to actually not link the bucket to the registry. Instead, we will link each bucket to a special type of process called Supervisors, which are explicitly designed to handle failures and crashes. We will learn more about them in the next chapter.
```

mfrasca:
(I did not double check the merge, hope it's ok) —
do you mean, the **two** supervisors you're talking about in the supervisors page, it's one for the buckets and one for the registry ?

josevalim:
I completely remove the two supervisors reference because we only implement the other supervisor in another chapter.

josevalim:
Thanks for all the feedback. I have pushed a bunch of text, examples, etc. I believe the chapter is much more complete now!

mfrasca:
reading flows like a charm, thank you!

PR #1280, opened 2019-06-13
different meanings of "application"

mfrasca:
the module, the function, the definition file.


josevalim:
I have pushed a correct definition. I believe you'd still prefer to define what is an application upfront but I think it makes more sense to define it after we seen the contents of the `.app` file.

josevalim:

mfrasca:
> I have pushed a correct definition. I believe you'd still prefer to define what is an application upfront but I think it makes more sense to define it after we seen the contents of the `.app` file.

well, yes, I prefer the structure I suggested. I am not even sure the three concepts are the ones.  there's also the `kv.ex` file, where we define the `KV` module, derived from `Application`. together with `application/0`, it's all automatically produced stuff, the user may not have a clue of their existence and content. unless one tells them.

let me insist on this point. before you start explaining "applications", mention what are the different possible concrete meanings of the term.

PR #1282, opened 2019-06-14
mention IEx.Helpers.recompile

mfrasca:
I think this is the correct place to mention it, since it's the first time we start iex inside the project.

### DIFF
--- a/getting-started/mix-otp/introduction-to-mix.markdown
+++ b/getting-started/mix-otp/introduction-to-mix.markdown
@@ -162,17 +162,17 @@ Once the project is compiled, you can start an `iex` session inside the project 
 $ iex -S mix
 ```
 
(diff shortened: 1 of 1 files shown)


josevalim:

mfrasca:
thanks for merging, but I have a perplexity here.
you removed the `()` after the function.
isn't this going to be deprecated soon?

josevalim:
Good point. iex allows to invoke the helpers without parens but I would add
them here to avoid confusion.
-- 


mfrasca:
you know, the 'mention a function evaluates it', that's what kept me away from ruby.

josevalim:
Yeah, in Elixir it only works without warnings inside IEx for convenience.

In any case, parens were added back in a more recent commit. :D

>

PR #1283, opened 2019-06-15
mention reason why :debugger may not be available

mfrasca:


josevalim:
I copied a note which we had in the dymamic supervisor chapter, thanks!

mfrasca:
yes, I just reached it.  looks good.

PR #1285, opened 2019-06-15
clarify hidden points

mfrasca:
for people coming from object oriented languages, these points are quite tricky indeed.

### DIFF
--- a/getting-started/mix-otp/dynamic-supervisor.markdown
+++ b/getting-started/mix-otp/dynamic-supervisor.markdown
@@ -52,7 +52,9 @@ We are going to solve this issue by defining a new supervisor that will spawn an
 
 ## The bucket supervisor
 
-Let's define a DynamicSupervisor and give it a name of `KV.BucketSupervisor`. This dynamic supervisor will be listed as a  child in our application supervisor. Replace the `init` function in `lib/kv/supervisor.ex` as follows:
+Since a `DynamicSupervisor` does not define any children during initialization, the `DynamicSupervisor` also allows us to skip the work of defining a whole separate module with the usual `start_link` function and the `init` callback. Instead, we can define a `DynamicSupervisor` directly in the supervision tree, by giving it a name and a strategy.
+
+Open up `lib/kv/supervisor.ex` and add the dynamic supervisor as a child as follows:
 
 
 ```elixir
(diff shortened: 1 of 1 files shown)


michalmuskala:
I wonder if it would maybe be easier to understand if regular atom was used as a name instead of the module-style one.

mfrasca:
which would be the equivalent form? using anything else than `KV.SupervisorBucket` would remove the advantage of having uniform, predictable values for the atoms used as process names.

michalmuskala:
You don't have this guarantee though. Anybody can use any atom as a name of any process.

mfrasca:
that's very clear. and since we are defining names ourselves, we are free to fix our naming logic. I think this part was made explicit, when the documentation introduces the reader to naming processes.

mfrasca:
or you would want to make it explicit in the supervisor-and-application page, near the first occurrence of `name:`.  but I don't think it would be necessary there.  I don't know.

mfrasca:
later on in the page, there's this sentence: 
> `DynamicSupervisor.start_child/2` expects the name of the supervisor and the child specification of the child to be started.

and I'm perplex about the first parameter, the "name of the supervisor".  also `h(DynamicSupervisor.start_child/2)` doesn't say anything about the `supervisor` parameter.  

can we use also the supervisor `pid`?  or it really needs be the name?

whatyouhide:
> also h(DynamicSupervisor.start_child/2) doesn't say anything about the supervisor parameter

Well, it links to the `Supervisor.supervisor/0` type which might be enough?

mfrasca:
> Well, it links to the `Supervisor.supervisor/0` type which might be enough?

to me: no.  it didn't say anything to me.  I was also not aware of this `type` concept in Elixir, nor had I noticed the `t` help function in `iex`.

the second argument is described as `Supervisor.child_spec()`, but the documentation text is more explicit on what you are expected to provide.

maybe I just missed the knowledge "look up type definitions, using `t`".

mfrasca:
(I'm afraid we aren't discussing the proposed change any more. my fault to open a side discussion.)

josevalim:
Thanks @mfrasca. I have removed some of the discussion to before we write the code and clarified the text.

josevalim:

PR #1287, opened 2019-06-17
the 2 things to notice is 4 according to me

mfrasca:
I split one of the points (define a module, define a test case) into three: define a module that uses ExUnit.Case, use **two** of the injected macros.

### DIFF
--- a/getting-started/mix-otp/introduction-to-mix.markdown
+++ b/getting-started/mix-otp/introduction-to-mix.markdown
@@ -189,11 +189,15 @@ defmodule KVTest do
 end
 ```
 
-It is important to note a couple things:
+It is important to note a couple of things:
 
 1. the test file is an Elixir script file (`.exs`). This is convenient because we don't need to compile test files before running them;
 
-2. we define a test module named `KVTest`, use [`ExUnit.Case`](https://hexdocs.pm/ex_unit/ExUnit.Case.html) to inject the testing API and define a simple test using the `test/2` macro;
+2. we define a test module named `KVTest`, in which we [`use ExUnit.Case`](https://hexdocs.pm/ex_unit/ExUnit.Case.html) to inject the testing API;
+
+3. we use one of the injected macros, `doctest` (TODO: for this I do not find documentation), to indicate that the `KV` module contains doctests (we are not going to discuss these further for the moment);
+
+4. we use the `text/2` macro to define a simple test;
 
 Mix also generated a file named `test/test_helper.exs` which is responsible for setting up the test framework:
 
--- a/getting-started/mix-otp/introduction-to-mix.markdown
+++ b/getting-started/mix-otp/introduction-to-mix.markdown
@@ -195,7 +195,7 @@ It is important to note a couple of things:
 
 2. we define a test module named `KVTest`, in which we [`use ExUnit.Case`](https://hexdocs.pm/ex_unit/ExUnit.Case.html) to inject the testing API;
 
-3. we use one of the injected macros, `doctest` (TODO: for this I do not find documentation), to indicate that the `KV` module contains doctests (we are not going to discuss these further for the moment);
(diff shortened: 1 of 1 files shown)


mfrasca:
I also have a question, regarding the `doctest` … is it possible to have something similar to Python's `ELLIPSIS`? I would expect to be able to write this, as a valid test case:

iex> KV.start(:ok, [])
{:error, {:already_started, _}}

but it tells me I can't put the _ here.

(CompileError) (for doctest at) lib/kv.ex:10: invalid use of _. "_" represents a value to be ignored in a pattern and cannot be used in expressions

josevalim:
re doctest: In Elixir you can do pattern matching:

```iex
iex> {:error, {:already_started, _}} = KV.start(:ok, [])
```

The doctest is rather a `==` comparison.

getting-started/mix-otp/introduction-to-mix.markdown
```diff
@@ -189,11 +189,15 @@ defmodule KVTest do
...
+It is important to note a couple of things:
 
 1. the test file is an Elixir script file (`.exs`). This is convenient because we don't need to compile test files before running them;
 
-2. we define a test module named `KVTest`, use [`ExUnit.Case`](https://hexdocs.pm/ex_unit/ExUnit.Case.html) to inject the testing API and define a simple test using the `test/2` macro;
+2. we define a test module named `KVTest`, in which we [`use ExUnit.Case`](https://hexdocs.pm/ex_unit/ExUnit.Case.html) to inject the testing API;
+
+3. we use one of the injected macros, `doctest` (TODO: for this I do not find documentation), to indicate that the `KV` module contains doctests (we are not going to discuss these further for the moment);
+
+4. we use the `text/2` macro to define a simple test;
```
josevalim:
```suggestion
4. we use the [`test/2`](https://hexdocs.pm/ex_unit/ExUnit.Case.html#test/3) macro to define a simple test;
```

getting-started/mix-otp/introduction-to-mix.markdown
```diff
@@ -189,11 +189,15 @@ defmodule KVTest do
...
 
-It is important to note a couple things:
+It is important to note a couple of things:
 
 1. the test file is an Elixir script file (`.exs`). This is convenient because we don't need to compile test files before running them;
 
-2. we define a test module named `KVTest`, use [`ExUnit.Case`](https://hexdocs.pm/ex_unit/ExUnit.Case.html) to inject the testing API and define a simple test using the `test/2` macro;
+2. we define a test module named `KVTest`, in which we [`use ExUnit.Case`](https://hexdocs.pm/ex_unit/ExUnit.Case.html) to inject the testing API;
+
+3. we use one of the injected macros, `doctest` (TODO: for this I do not find documentation), to indicate that the `KV` module contains doctests (we are not going to discuss these further for the moment);
```
josevalim:
```suggestion
3. we use one of the injected macros, [`doctest/1`](https://hexdocs.pm/ex_unit/ExUnit.DocTest.html#doctest/2), to indicate that the `KV` module contains doctests (we will discuss those in a later chapter);
```

mfrasca:
Re: doctest:
I had done that too, the pattern match, but still I get a result which I do not know how to test for correctness, since it includes the `pid` which can be anything.

mfrasca:
(short of doing something like this, which defeats the goal of doctesting, I think)
```iex
      iex> KV.start(:ok, []) |> (fn({a, {b, _}}) -> {a, b} end).()
      {:error, :already_started}
```

josevalim:
@mfrasca you don't need to do anything. Just this should be enough:

    iex> {:error, {:already_started, _}} = KV.start(:ok, [])

josevalim:

mfrasca:
> @mfrasca you don't need to do anything. Just this should be enough:
> 
> ```
> iex> {:error, {:already_started, _}} = KV.start(:ok, [])
> ```

I'm not making myself understood.  I want to write a valid `doctest`.

josevalim:
Oh, I thought we could skip the comparison part. Hrm, sorry.

PR #1298, opened 2019-06-26
Add Slackware install instructions

YellowApple:
I've only been maintaining these SlackBuilds for 3 years now; might as well advertise 'em :)

Note that at the moment SlackBuilds.org does not have the latest versions of Erlang/Elixir (my bad), but I submitted those today, so they'll hopefully be approved and included in the next SBo update cycle.  The updated versions of the SlackBuild scripts can be found in [the GitHub repo I use for maintaining my SlackBuilds](https://github.com/YellowApple/slackbuilds).


josevalim:

PR #1299, opened 2019-07-04
Supervisor section changes

RabeaGleissner:
I was going through the Supervisor tutorial (it’s fantastic! Thank you so much for that!) and noticed a few small inconsistencies along the way.

I made the following changes:

* Removed `extra_applications` from the `.app` file code - I checked on Elixir 1.8.1. and it's not there anymore... Correct me if I'm wrong, of course!
* Made variable naming consistent in the implementation of `handle_cast`. Sometimes the second element of the tuple that is returned from `KV.Bucket.start_link([])` was called `bucket` and sometimes `pid`. I changed it to `bucket`.
* Corrected minor typos
  - "what happens when something fail?" --> "what happens with something fail`s`?"
  - "because a certain condition has reached" --> "because a certain condition `w`as reached"
  - "But how we can automatically start the supervisor" --> "But `how can we` automatically start the supervisor"

Please let me know if this makes sense. Happy to make any changes to these changes 😄


josevalim:

PR #1300, opened 2019-07-07
add installation instructions for OpenBSD

jasperla:
OpenBSD has been providing elixir packages since 0.5.0 so we might as well list it here.


josevalim:

PR #1308, opened 2019-07-25
Remove Code Elixir LDN 2019 from upcoming conferences

gausby:
The Code Elixir LDN 2019 conference is over, so I have removed it from the list of important links.


josevalim:

PR #1311, opened 2019-08-03
Multi headed anon function docs

chriseyre2000:
I could not find an example of multi-headed anonymous functions in the official documentation.

One of my students on the Elixir track of Exercism.io was looking for the docs on this. 
I promised to submit some.


josevalim:
Thanks @chriseyre2000! However, I don't think it make sense for us to talk about it this early on. at this moment, we have barely introduced pattern matching, so the idea of multiple clauses can be confusing.

Maybe we can talk about it when talking about multiple clauses in functions.

The other point is that the guides are not meant to be a complete reference, so we it will always be missing informatation. Perhaps this would be better documented in the `Function` module docs?

vsyrovat:
@chriseyre2000 thnx for interesting syntax example!

josevalim:
Closing this as a stale. We are open to suggestions to move this elsewhere that is not so early. Thank you!

PR #1316, opened 2019-08-13
Minor documentation fix

mjmostachetti:
I think these are mixed up.


mjmostachetti:
On second though, I think the original is right. The wording was confusing me.

PR #1317, opened 2019-08-18
Update example in Enumerables and Streams docs

vlttnv:
When going through the tutorial I noticed that when executing `stream = File.stream!("path/to/file")` it does not return a function like in the docs `#Function<18.16982430/2 in Stream.resource/3>`.
In the proposed change I have added what was returned by `iex`.

I am not sure if tha was a mistake. Please close the PR if the docs are correct.


josevalim:

PR #1326, opened 2019-09-05

aruprakshit:
Replace outdated error message with new error message text while using macro without require.


josevalim:

PR #1328, opened 2019-10-02
Update ets.markdown with accessibility example

vysogot:
While playing with ETS I was getting a generic ArgumentError. I thought I'm passing the arguments to the `:ets.insert` incorrectly, while the real cause was the access setting of the ETS table. I did not this information anywhere on `elixing-lang.org` and I believe it should be here.

### DIFF
--- a/getting-started/mix-otp/ets.markdown
+++ b/getting-started/mix-otp/ets.markdown
@@ -28,7 +28,13 @@ iex> :ets.lookup(table, "foo")
 [{"foo", #PID<0.41.0>}]
 ```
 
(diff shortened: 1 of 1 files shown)


josevalim:
Thanks Jakub. I am a bit worried that adding this information later will be
confusing. People will wonder if they have to change the visibility or not.
Therefore I have a proposal: what if we pass the visibility when we first
create it? Then we can discuss about the possible values, the default, and
why we picked it. What do you think?
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

vysogot:
Hello José, thanks for your feedback. I changed it. Would it be less confusing now?

josevalim:

PR #1334, opened 2019-10-13
The name of the file was mentioned that way before

JJ:
It's not a script, so it makes sense to use it that way too.

### DIFF
--- a/getting-started/modules-and-functions.markdown
+++ b/getting-started/modules-and-functions.markdown
@@ -276,7 +276,7 @@ If we save the code above in a file named "concat.ex" and compile it, Elixir wil
 The compiler is telling us that invoking the `join` function with two arguments will always choose the first definition of `join` whereas the second one will only be invoked when three arguments are passed:
 
 ```console
-$ iex concat.exs
+$ iex concat.ex
 ```
 
 ```iex


josevalim:
Given we have Elixir code at the end of the file, I would rather name it a
script. Perhaps we should be clear that Concat should be written to a
concat.exs file when we first create it?
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

JJ:
Whatever you do, it should be coherent... I chose .ex because it was the
way it was mentioned before. Maybe I should have changed that to .exs

josevalim:
Do we mention concat.ex anywhere else
In the article though? Do you want to change it when we first create it so
it is clear it is a .exs file?
-- 


JJ:
It's right there, in line 272:

```
If we save the code above in a file named "concat.ex" and compile it, Elixir will emit the following warning:
```

I do think this is a concat.ex; it's a simple module (defined in lines 258 to 270), not a script.

josevalim:

josevalim:
I missed it, thanks for following up!

JJ:
No problem, my pleasure!

PR #1338, opened 2019-11-01
x=1 declaration not clear in pattern-match readme

sanchitbansal10:
The code is assuming that x = 1 is declared and then match is checked.
For a new user it's not very clear. I read it for the first time (completely new to Elixir) and thought that 1 = x is the declaration.


josevalim:

PR #1342, opened 2019-11-28
Add upcoming conferences section and add CodeBeam SF & ElixirConf EU

gausby:
Added in important links; over the community section.

I have made it such that it is easy to add and remove conferences from a list; I thought it would be best to order the conferences such that the most recent is at the top, and the template includes room for a logo, the location (city, country/state), and a little space to write some information about dates. Hopefully this will make it easy to add and remove conferences from the list.


josevalim:

PR #1346, opened 2019-12-05
doc: Clarify available functions in attribute definitions

rawtaz:
Clarifies that functions in the same module as the attribute is defined in cannot be called as they are not compiled yet.


josevalim:

PR #1349, opened 2019-12-13
Update module-attributes.markdown

fireproofsocks:
I wanted to clarify a few things:
-  module attributes are not constants -- it really confused me when I had "constants" locked in my brain, and then I saw that they could be reassigned.
- Point out ExUnit's use of the `@tag` module attribute a bit earlier.
- I wanted to point out the "accumulate" option
- I wanted to see an example of calling functions when a module attribute is defined.


josevalim:

PR #1350, opened 2019-12-13

fireproofsocks:
I wanted to show an example of how to inspect a string without modifying it.


josevalim:

PR #1351, opened 2019-12-15
Updating Documentation on Config and Releases

2dwn:
* '--config' flag is deprecated
* received 'Connection attempt from disallowed node :"foo@computer-name"' when the cookie option was not set on the releases


josevalim:
Good catch @2dwn. I decided to remove that section altogether because it doesn't seem to be relevant anymore, especially because we can do these different configurations inside the release.

josevalim:

PR #1353, opened 2019-12-19

fireproofsocks:
Clarifying that only public functions may be imported.


josevalim:
Thank you! :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #1356, opened 2020-01-15
Added age key to Map.put example

mortalisk:
The age key seemed to come from nowhere in this example. Added it so that the example is correct.


josevalim:
The age is defined as a default value for the struct here:

https://github.com/elixir-lang/elixir-lang.github.com/pull/1356/files#diff-3f058a769478cf6679fe0fa97b5370bcR30

In any case, thanks for the PR!

PR #1361, opened 2020-01-29
Update protocols.markdown

fireproofsocks:
Begins the explanation by exploring a more familiar protocol: `String.Chars`; includes an example of implementing `String.Chars` for a custom data type.

### DIFF
--- a/getting-started/protocols.markdown
+++ b/getting-started/protocols.markdown
@@ -7,9 +7,50 @@ title: Protocols
 
 {% include toc.html %}
 
-Protocols are a mechanism to achieve polymorphism in Elixir. Dispatching on a protocol is available to any data type as long as it implements the protocol. Let's see an example.
+Protocols are a mechanism to achieve polymorphism in Elixir; by implementing the functions specified by a protocol, data types can define functionality appropriate to certain situations. Dispatching on a protocol is available to any data type that has implemented the protocol. One common example that you are probably familiar with is the seemingly innocuous "magic" that handles converting variables to binaries during string interpolation:
 
-In Elixir, we have two idioms for checking how many items there are in a data structure: `length` and `size`. `length` means the information must be computed. For example, `length(list)` needs to traverse the whole list to calculate its length. On the other hand, `tuple_size(tuple)` and `byte_size(binary)` do not depend on the tuple and binary size as the size information is pre-computed in the data structure.
+```iex
+iex> n = 0
+0
+iex> "I got #{n} problems."
+"I got 0 problems."
+```
+
+We can interpolate a variable containing an integer into a string because Elixir includes a built-in implementation of the `String.Chars` protocol which calls the `Integer.to_string/1` function behind the scenes:
+
+```elixir
+defimpl String.Chars, for: Integer do
+  def to_string(term) do
+    Integer.to_string(term)
+  end
+end
+```
+We can infer that the `String.Chars` protocol is responsible for converting a structure to a binary (when applicable), but what does the protocol definition actually look like? A protocol definition makes use of the special `defprotocol` keyword, and it uses function specs and stubs in a way that may look similar to how other languages define interfaces or abstract base classes:
+
+```elixir
+defprotocol String.Chars do
+  @spec to_string(t) :: String.t()
+  def to_string(term)
+end
+```
+
+Implementing the `String.Chars` protocol requires that you implement a single function: `to_string/1`. 
+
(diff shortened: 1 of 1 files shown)


josevalim:
Thanks @fireproofsocks. I think starting with the interpolation example introduces a couple questions which may cause more confusion in the long term. In particular:

  1. The first implementation you show is for a struct which is only introduced in the second section. So it feels a bit less organized.
  2. What is an ObjectID? What if I am not familiar with mongo?
  3. Why is interpolation not automatically implemented for PID and other Elixir data types? Should I implement them?

So while the problem is simpler than size/length, I think it introduces complexities in other areas which I am not sure are worth it.

fireproofsocks:
Hmm... the particular struct demonstrated wasn't really the point there, but I think the approach could be improved even more.  One of the pieces that I feel is missing from the docs in their current state is that they are not starting with a familiar concept (like `to_string`), and secondly, I don't think the current examples make it clear enough HOW the implementations get called.

How would you feel if the introduction took a slightly different tack and discussed protocols/implementations in the context of the "Open-Closed Principle"?  For me (and the devs I've tried to explain this to), it always feels like the current docs aren't quite crossing the bridge there.  

Since pattern matching has already been discussed, what if the opening example was a variation on `to_string`?  E.g. if you had to write code that would covert different data types to a "masked" string, you might come up with something like this:

```elixir
defmodule Mystrings do
  def mask(value) when is_binary(value) do
    "--string--"
  end

  def mask(value) when is_integer(value) do
    "--integer--"
  end
  # ... other functions/guards to handle other data types ...
end
```
where the output would be something like:
```iex
iex> Mystrings.mask("foo")
"--string--"
iex> Mystrings.mask(123)
"--integer--"
```
Developers should be able to see how they could implement more functions using guard clauses or pattern matching to support more data types.  And if they continue that line of reasoning, that should help make it clear the problem that protocols/implementations help solve: you can't expect everyone using your code to be able to add functions to your modules ad infinitum -- your code should be closed to modifications, but open to extension.

So that's when you could demonstrate the protocol/implementation variation of the above:
` ``elixir
defprotocol Mystrings do
  @spec mask(t) :: String.t()
  def mask(value)
end

defimpl Mystrings, for: BitString do
  def mask(_value) do
    "--string--"
  end
end

defimpl Mystrings, for: Integer do
  def mask(_value) do
    "--integer--"
  end
end
# ... more implementations for other data types...
```
and it would have the identical output, but that keeps the original code closed to modification, but anyone can add on implementations of the protocol.  

There may have to be some clarifications on the `BitString` there, but I think the idea of showing this alongside an alternative, stripped of any detailed functionality, should be more educational.

Thoughts?

josevalim:
Agreed! Talking about the limitations of pattern matching and how protocols
address them is important. It could definitely be used to replace the
current intro. I would not explicitly mention the open/closed principle
though, as many may not be familiar with it (and those who are will
recognize it).
-- 


*José Valimhttps://dashbit.co/ <https://dashbit.co/>*

fireproofsocks:
I think it would be a great teachable moment to mention the Open/Closed principle because it's specifically soooo relevant there.  Just mention it as an aside with a link, and the lightbulb will go off for someone because they will see it in action.  In my professional opinion as an instructor, that is the PERFECT place to mention that concept.

The final thing that I think is missing from this is another thing that has continuously confused me and plagued my ability to explain these Elixir concepts to members of my team.  Protocols/implementations are often mentioned as being Elixir's take on "interfaces"... and sometimes people say the same thing about behaviours, so it gets confusing...

**The docs need to make a clarification between protocols/implementations and behaviours**.  I.e. when should one be used over the other?  Specifically, protocols/implementations are appropriate when the functionality must be different _depending on the data type_.  Behaviours are more flexible in that their implementations may all deal with the same data-types.  Correct me if I'm wrong, but something like `Ecto` queries must rely on behaviours and not protocols, right?  An Ecto query gets dispatched that a database adapter (e.g. one specific for Postgres or MySQL).  A protocol wouldn't work in that case because the data-type of the query variable is the same in all cases.  Right?

Anyhow, I think that covers the last of my concerns.  Thoughts?

josevalim:
The getting started guides are meant to be introductory guides. I understand it can be beneficial to provide some references as we go but we have generally refused that over time because otherwise the guides would have become very dense. I prefer to keep the guides "fun" and light-weight.

Regarding protocols and behaviours, we could talk about it, but on chapter 20, when we introduce behaviours. Behaviours have not yet been introduced at this point.

Ecto uses both protocols and adapters. When you do `from p in something`, something uses the queryable protocol. `Repo.all(Post)` uses the same protocol too. But the adapters are indeed behaviours. :)

Thanks for the discussion!

fireproofsocks:
How about I send up a separate PR?

josevalim:
Sounds great!

fireproofsocks:
Submitting PR #1365 as an alternative to this PR.

josevalim:
Fantastic work, thank you!

PR #1362, opened 2020-02-03
Update the URI to the ElixirConf EU logo

gausby:
The image file for the ElixirConf EU logo in the upcoming conferences section had gone awol. I've updated the uri to one that will hopefully stay up *forever*!


josevalim:

Issue #1172, opened 2018-09-10
Mix and OTP guide: mix test does not run in "Doctests, patterns and with"


FumioNonaka:
https://github.com/elixir-lang/elixir-lang.github.com/edit/master/getting-started/mix-otp/docs-tests-and-with.markdown#L73
The following error is shown:
```
** (Mix) Could not start application kv_server: exited in: KVServer.Application.start(:normal, [])
    ** (EXIT) an exception was raised:
        ** (RuntimeError) missing $PORT environment variable
```
It is caused by the code below in the `KVServer.Application.start/2`.
```
port = String.to_integer(System.get_env("PORT") || raise "missing $PORT environment variable")
```
I changed the line into:
```
port = String.to_integer(System.get_env("PORT") || "4040")
```

fidenz-chim:
Added this modification to the documentation.

PR #1367, opened 2020-02-16
Fix inconsistency regarding default port env var

Hoffs:
Very similar to #1172 which missed last rewrite part
that still raised error if env variable was not present.

So if someone were to just copy last part then in next section tests would fail when doing simple `mix test`.


josevalim:

PR #1368, opened 2020-02-16
Correct the Arity of 'call' in the ETS Guide

nskins:
In the ETS guide, the function is written as `call/3`, but when we fix the issue in code, we use `call/2`. This is also supported by other written references to `call/2`.


josevalim:

PR #1374, opened 2020-03-04

fireproofsocks:
This represents a substantial rewrite of this page that seeks to address some of the frequent misunderstandings that arise when discussing strings and binaries. Many of the explanations were initially given in this forum post: https://elixirforum.com/t/where-did-the-name-binaries-come-from-and-how-does-this-relate-to-base2/29490/16 -- many thanks to the wise input shared by contributors there.

This change:
- normalizes the spelling of "code point" (per Wikipedia), using a space (and not "codepoint")
- organizes the explanations to sensibly move from Unicode, to code points, to encodings, to bitstrings, to binaries, to strings (in that order) 
- adds explanations and examples of the `::size(n)`, `::binary-size(n)` and `::utf8` modifiers
- adds more examples
- provides an explanation as to why Elixir refers to strings as "binaries"

Since this changes so much of the existing text, I would recommend reviewing the new version of the file as a whole to check it for readability and accuracy: https://github.com/elixir-lang/elixir-lang.github.com/blob/e4b01dfe5fe326fc9e78640b6a6edb106f212ca2/getting-started/binaries-strings-and-char-lists.markdown

I'm not sure how the `\\u` will render... In ExDoc comments I've had to double-escape them, but I'm not sure if that's the case here.

### DIFF
--- a/getting-started/binaries-strings-and-char-lists.markdown
+++ b/getting-started/binaries-strings-and-char-lists.markdown
@@ -35,7 +35,7 @@ iex> ?ł
 
 It's like a DNS lookup for characters to find their Unicode code point!
 
-Note that most Unicode code charts will refer to a code point by its hexadecimal representation, e.g. `97` translates to `0061` in hex, and we can represent any Unicode character in an Elixir string by using the `\\u` notation and the hex representation of its code point number:
+Note that most Unicode code charts will refer to a code point by its hexadecimal representation, e.g. `97` translates to `0061` in hex, and we can represent any Unicode character in an Elixir string by using the `\u` notation and the hex representation of its code point number:
 
 ```iex
 iex> "\\u0061" === "a"
--- a/getting-started/binaries-strings-and-char-lists.markdown
+++ b/getting-started/binaries-strings-and-char-lists.markdown
@@ -38,7 +38,7 @@ It's like a DNS lookup for characters to find their Unicode code point!
 Note that most Unicode code charts will refer to a code point by its hexadecimal representation, e.g. `97` translates to `0061` in hex, and we can represent any Unicode character in an Elixir string by using the `\u` notation and the hex representation of its code point number:
 
 ```iex
-iex> "\\u0061" === "a"
+iex> "\u0061" === "a"
 true
 ```
 
--- a/getting-started/binaries-strings-and-char-lists.markdown
+++ b/getting-started/binaries-strings-and-char-lists.markdown
@@ -7,7 +7,7 @@ title: Binaries, strings, and charlists
 
 {% include toc.html %}
 
-In "Basic types", we learned about strings and used the `is_binary/1` function for checks:
+In "Basic types", we learned a little bit about strings and we used the `is_binary/1` function for checks:
 
 ```iex
 iex> string = "hello"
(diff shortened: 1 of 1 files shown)


getting-started/binaries-strings-and-char-lists.markdown
```diff
@@ -50,45 +33,43 @@ iex> ?ł
 322
 ```
 
-These can be used anywhere you want to refer to a characters codepoint.
+It's like a DNS lookup for characters to find their Unicode code point!
```
josevalim: I like this metaphor but I am afraid that, if you don't know DNS lookups, then you will leave more confusing. We are light on metaphor on the guides for this reason, so I would recommend to remove it. :)

getting-started/binaries-strings-and-char-lists.markdown
```diff
@@ -16,32 +16,15 @@ iex> is_binary(string)
...
+In this chapter, we will gain clarity on what exactly binaries are, how they relate to strings, and what single-quoted values, `'like this'`, mean in Elixir. Although strings are one of the most common data types in computer languages, they are subtly complex and are often misunderstood. To understand strings in Elixir, we have to educate ourselves about [Unicode](https://en.wikipedia.org/wiki/Unicode) and character encodings, specifically the [UTF-8](https://en.wikipedia.org/wiki/UTF-8) encoding.
 
-## UTF-8 and Unicode
+## Unicode and Code Points
 
-A string is a UTF-8 encoded binary. In order to understand exactly what we mean by that, we need to understand the difference between bytes and code points.
+In order to facilitate meaningful communication between computers across multiple languages, a standard is required so that the ones and zeros on one machine mean the same thing when they are transmitted to another. The [Unicode Standard](https://unicode.org/standard/standard.html) acts as an official registry of virtually all the characters we know: this includes characters from classical and historical texts, emoji, and formatting and control characters as well.
```
josevalim: Great intro!

getting-started/binaries-strings-and-char-lists.markdown
```diff
@@ -50,45 +33,43 @@ iex> ?ł
 322
 ```
 
-These can be used anywhere you want to refer to a characters codepoint.
+It's like a DNS lookup for characters to find their Unicode code point!
 
-You can also use the functions in [the `String` module](https://hexdocs.pm/elixir/String.html) to split a string in its individual characters, each one as a string of length 1:
+Note that most Unicode code charts will refer to a code point by its hexadecimal representation, e.g. `97` translates to `0061` in hex, and we can represent any Unicode character in an Elixir string by using the `\\u` notation and the hex representation of its code point number:
```
josevalim:
```suggestion
Note that most Unicode code charts will refer to a code point by its hexadecimal representation, e.g. `97` translates to `0061` in hex, and we can represent any Unicode character in an Elixir string by using the `\u` notation and the hex representation of its code point number:
```

getting-started/binaries-strings-and-char-lists.markdown
```diff
@@ -50,45 +33,43 @@ iex> ?ł
...
 
 ```iex
-iex> String.codepoints("hełło")
-["h", "e", "ł", "ł", "o"]
+iex> "\\u0061" === "a"
```
josevalim:
```suggestion
iex> "\u0061" === "a"
```

getting-started/binaries-strings-and-char-lists.markdown
fireproofsocks: Ok, removed.

getting-started/binaries-strings-and-char-lists.markdown
fireproofsocks: Thanks!

josevalim:
Absolutely fantastic work!  I have just simplified the introduction to UTF-8 to skip the part about fixed and variable width and jump straight to the point. More improvements and suggestions are definitely welcome!

josevalim:

everettvody:
Awesome, thank you! I hope these clarifications will help educate others. After so many years, I *finally* understand encodings.

fireproofsocks:
Awesome, thank you! I hope these clarifications will help educate others. After so many years, I finally understand encodings.

EskiMag:
This was super useful. I clarified quite a lot of things thanks to this PR. Thanks a lot @fireproofsocks!

PR #1375, opened 2020-03-06

fireproofsocks:
A few general cleanups -- better demo of what happens when you match the first byte of a multibyte character.  Some polish to phrasing here and there.


josevalim:

PR #1379, opened 2020-04-10
Update pattern-matching.markdown

fireproofsocks:
Adds a few clarifications and examples re the pin operator.

### DIFF
--- a/getting-started/pattern-matching.markdown
+++ b/getting-started/pattern-matching.markdown
@@ -148,7 +148,8 @@ iex> 1 = 2
 Notice that we even see the exact same error message.
 
 We can use the pin operator inside other pattern matches, such as tuples or lists:
-```
+
+```iex
 iex> x = 1
 1
 iex> [^x, 2, 3] = [1, 2, 3]
--- a/getting-started/pattern-matching.markdown
+++ b/getting-started/pattern-matching.markdown
@@ -171,7 +171,7 @@ iex> {y, 1} = {2, 2}
 ** (MatchError) no match of right hand side value: {2, 2}
 ```
 
-There is one caveat when using the pin operator in maps: you must use the `=>` syntax when pinning a map's key, even when the value being pinned is an atom:
+When using the pin operator in maps you must use the `=>` syntax when pinning a map's key, even when the value being pinned is an atom:
 
 ```iex
 iex> k = :foo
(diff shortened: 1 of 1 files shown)


getting-started/pattern-matching.markdown
```diff
@@ -129,14 +129,30 @@ iex> x = 1
...
+Because we have pinned `x` when it was bound to the value of `1`, it is equivalent to the following:
+
+```iex
+iex> 1 = 2
+** (MatchError) no match of right hand side value: 2
+```
+Notice that we even see the exact same error message.
+
+We can use the pin operator inside other pattern matches, such as tuples or lists:
+```
```
fertapric, line 151: Nitpick:

```suggestion
We can use the pin operator inside other pattern matches, such as tuples or lists:

```iex
```

getting-started/pattern-matching.markdown
```diff
@@ -145,13 +164,26 @@ iex> {y, ^x} = {2, 2}
...
-Because we have assigned the value of 1 to the variable x, this last example could also have been written as:
+Because `x` was bound to the value of `1` when it was pinned, this last example could have been written as:
 
-```
+```iex
 iex> {y, 1} = {2, 2}
 ** (MatchError) no match of right hand side value: {2, 2}
 ```
 
+There is one caveat when using the pin operator in maps: you must use the `=>` syntax when pinning a map's key, even when the value being pinned is an atom:
```
fertapric, line 174:
```suggestion
When using the pin operator in maps you must use the `=>` syntax when pinning a map's key, even when the value being pinned is an atom:
```

getting-started/pattern-matching.markdown
```diff
@@ -129,14 +129,33 @@ iex> x = 1
 iex> x = 2
 2
 ```
+However, there are times when we don't want variables to be rebound.
 
-Use the pin operator `^` when you want to pattern match against an existing variable's value rather than rebinding the variable:
+**Use the pin operator `^` when you want to pattern match against a variable's _existing value_ rather than rebinding the variable.**
```
josevalim, line 134: We don't use bold anywhere else in the guides, so I wouldn't start now. Everything in the guide is important, otherwise it shouldn't be in the guide. :)

josevalim:
Ping. :)

getting-started/pattern-matching.markdown
fireproofsocks:
```suggestion
Use the pin operator `^` when you want to pattern match against a variable's _existing value_ rather than rebinding the variable.
```

fireproofsocks:
I approved suggestions and resolved conversations -- did I miss something?

josevalim:

PR #1382, opened 2020-04-15
Fixed typo: installitino -> installation

KevinGreene:


fertapric:
Thanks @KevinGreene! 💜

PR #1383, opened 2020-04-22
Update agent.markdown

LiquidSolid:
### Reason
The code above uses `Agent.update/2`, not `Agent.update/3`.

### Code sample
```elixir
@doc """
Puts the `value` for the given `key` in the `bucket`.
"""
def put(bucket, key, value) do
  # Only 2 arguments, yeah?
  Agent.update(bucket, &Map.put(&1, key, value))
end
```

### Note
**Maybe I'm wrong.**


josevalim:

PR #1385, opened 2020-04-22
Simplify example in enumerables-and-streams.markdown

Nagasaki45:
The `total_sum` variable is unnecessary for this example, and is not discussed in the text. Also, the example is later compared to an equivalent code snippet without the pipe operator. In the comparison code there's no `total_sum` so they are not exact equivalents, hence confusing. Lastly, the example already introduces a few new concepts, so it might be better to keep it simple. For these reasons, I think it's better to drop the `total_sum` variable.


josevalim:

PR #1386, opened 2020-04-24
add {max-width: 100%} css rule to all img tags to improve mobile UX

stelabrego:
I was following the guides and noticed the Debugging page has horizontal overflow on screens < 640px due to the images having a hardcoded width, so adding this CSS rule makes the images contained to the screen width, improving mobile UX.

You can see the problem is especially relevant on small mobile devices like the iPhone SE.

Before:
before

After:
after


josevalim:

PR #1388, opened 2020-04-26
Correct GenServer cheatsheet some 'client' blocks to 'callback's

Nagasaki45:
`handle_info` and `handle_continue` were described as client side functions. They are callbacks, as `handle_call` etc. Isn't it?


josevalim:

Issue #1387, opened 2020-04-24
Problem with distributed tasks tutorial


haffla:
Hello! I am working through the OTP tutorial and I think there is a problem in the [distributed tasks chapter]((https://elixir-lang.org/getting-started/mix-otp/distributed-tasks.html#wiring-it-all-up)). When I follow the instructions to use `KV.Router.route/4` in order to look up the process that holds a specific bucket (like "shopping") one of the tests written in a previous chapter breaks.

```elixir
# apps/ks_server/test/kv_server_test.exs
# ...
assert send_and_recv(socket, "CREATE shopping\r\n") == "OK\r\n"

# This one breaks. It returns "NOT FOUND"
assert send_and_recv(socket, "PUT shopping eggs 3\r\n") == "OK\r\n"
# ...
```

I am quite new to Elixir/Erlang/OTP and I am not sure if I missed some step in the tutorial. But isn't this expected? I've been reading up on ets and isn't it per node, so not distributed? The first request (CREATE) goes to the first node and the second (PUT) goes to the second node. Since on the second node "shopping" does not exist it returns "NOT FOUND". Did I miss a step that made ets distributed?

Anyway thanks for helping out and for this great tutorial.

josevalim:
You are correct, ets is not distributed. So you would have to route both create and put statements, so they all go the appropriate node. If it helps, you can find a working application here: github.com/josevalim/kv_umbrella

haffla:
@josevalim thanks. OK that's what I thought. I had a look at your kv_umbrella repo. You are right that the tests are green there but only when you have a routing table like `[{?a..?z, node()}]`. It's just one node so all request are going to one node. Change it to two nodes as described in this [section](https://elixir-lang.org/getting-started/mix-otp/distributed-tasks.html) and you will get the same failing test as described above.

I found the problem: https://github.com/josevalim/kv_umbrella/blob/master/apps/kv_server/lib/kv_server/command.ex#L48

The create command always runs on the current node. Would you accept a PR that addresses this?

This is the fix (`command.ex`)

```elixir
  def run({:create, bucket}) do
    case KV.Router.route(bucket, KV.Registry, :create, [KV.Registry, bucket]) do
      pid when is_pid(pid) -> {:ok, "OK\r\n"}
      _ -> {:error, "FAILED TO CREATE BUCKET"}
    end
  end

```

josevalim:
Yes, a PR would be very welcome. Thanks! :)

PR #1390, opened 2020-04-28
MIX/OTP Guide: Route create command to appropriate node

haffla:
This fixes an issue with the OTP guide which is best described in the issue itself.

https://github.com/elixir-lang/elixir-lang.github.com/issues/1387


josevalim:

PR #1393, opened 2020-05-02
Fix inconsistency in task-and-gen-tcp.markdown

wirelyre:
616d78c changed the example to use a default port if none was provided.
be1471b updated another code block to match that change.

This change fixes the remaining copy of that code.

Thus the trilogy of #1172 and #1367 is complete. :-)


josevalim:

PR #1397, opened 2020-05-17
Add note on multiple clauses in anonymous functions

rawtaz:
Adds a paragraph with code example under the "Anonymous functions" heading in "Basic types" in the Guide, explaining that multiple clauses can be specified with anonymous functions.


josevalim:
It is too early to introduce this here, as we haven't talked about pattern matching yet. We do talk about anonymous functions with multiple clauses in https://elixir-lang.org/getting-started/case-cond-and-if.html though. Thanks!

rawtaz:
Quite an unexpected place to find it in, and not a place you'd look for it in, if you were to want to find it in the Guide. That's why I filed the PR. But this just means we'll have more chat going on in #elixir-lang on freenode, which is of course good ;)

PR #1399, opened 2020-05-20
Interactive Elixir is only aliased on PowerShell.

CircuitRCAY:


CircuitCodes:
Both `iex` and `iex.bat` work perfectly fine in Windows, just that the `iex` command is aliased as a PowerShell command, which does _not_ carry over in Command Prompt.

josevalim:

PR #1401, opened 2020-05-28
fixup example explanation

mpisanko:
from the Docs: 
```
When registering an attribute, two options can be given:

    :accumulate - several calls to the same attribute will accumulate instead of overriding the previous one. New attributes are always added to the top of the accumulated list.
```


josevalim:

PR #1406, opened 2020-06-20
macros.markdown: add `()` to nullary macros

henrik:
As they would result in warnings otherwise.

### DIFF
--- a/getting-started/meta/macros.markdown
+++ b/getting-started/meta/macros.markdown
@@ -126,7 +126,7 @@ defmodule HygieneTest do
   def go do
     require Hygiene
     a = 13
-    Hygiene.no_interference
+    Hygiene.no_interference()
     a
   end
 end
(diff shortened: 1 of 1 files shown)


josevalim:
Those do not warn BUT it is nice to fix them anyway, thanks! 

josevalim:

henrik:
❤️ I tried copying one into my code and calling without parentheses in a test and got a warning, but maybe I missed something :)

henrik:
Checked again for my own sanity :D 

```
defmodule Lab do
  defmacro mymacro, do: IO.puts("Hello from Elixir #{System.version()}!")
end

defmodule Run do
  import Lab

  def run do
    mymacro
  end
end

Run.run
```

gets me

```
warning: variable "mymacro" does not exist and is being expanded to "mymacro()", please use parentheses to remove the ambiguity or change the variable name
  nofile:9: Run.run/0

Hello from Elixir 1.10.3!
```

josevalim:
Right, but those are qualified calls. `Lab.mymacro` doesn't warn, `mymacro` does. :)

henrik:
Ah! Glad I asked :) Thanks, José!

PR #1408, opened 2020-06-22
Add Groxio resources to elixir-lang.org/learning

batate:
Adding the two Groxio/Pragpub Elixir courses. OTP and Elixir.


josevalim:
Hi @batate! 

I am wondering if in cases like yours and PragProg, that have multiple courses, if it may be better to have a single entry for all of them. BUT, given that we have only three courses listed at the moment, this is a problem we can gladly save for later.

josevalim:

batate:
> Hi @batate!
> 
> I am wondering if in cases like yours and PragProg, that have multiple courses, if it may be better to have a single entry for all of them. BUT, given that we have only three courses listed at the moment, this is a problem we can gladly save for later.

Good point. I don't envision any more base Elixir courses. We'll see how these do. If there's interest, there are potential plans for Ecto, Nerves, and other projects that live elsewhere.

PR #1410, opened 2020-06-25
Adds missing "be"

gabrielmadruga:


getting-started/mix-otp/config-and-releases.markdown
```diff
@@ -107,7 +107,7 @@ Note we removed `async: true` from `use ExUnit.Case`. Since the application envi
 
 ## Releases
 
-Now that our application runs distributed, you may wondering how can we package our application to run in production. After all, all of our code so far depends on Erlang and Elixir versions that are installed in your current system. To achieve this goal, Elixir provides releases.
+Now that our application runs distributed, you might be wondering how we can package our application to run in production. After all, all of our code so far depends on Erlang and Elixir versions that are installed in your current system. To achieve this goal, Elixir provides releases.
```
eksperimental, line 110:
```suggestion
Now that our application runs distributed, you may be wondering how we can package our application to run in production. After all, all of our code so far depends on Erlang and Elixir versions that are installed in your current system. To achieve this goal, Elixir provides releases.
```

eksperimental: May is in present, might in past.

gabrielmadruga: @eksperimental I accepted your suggestion but I would like to have a source on the use of might for past and may for present. All I can find is that may is used to express greater possibility.
Thanks.

josevalim:

PR #1421, opened 2020-07-12
Update basic-types.markdown

ream88:
My nephew who is new to Elixir had problems understanding the string interpolation example because it was too dense. I split it into two lines to make it easier to understand.


josevalim:

PR #1422, opened 2020-07-13
use Application.compile_env

fuelen:
instead of Application.get_env


josevalim:
I think this example should actually be discouraged in general, so I picked something simpler. Thanks!

josevalim:

PR #1423, opened 2020-07-15
paths for kv application and quotes around atoms

dischoen:
Hi,

I just ran through getting-started and noticed some inconsistencies with current elixir (1.10.2 on erlang22).
These are:
* paths for the source files are specified the other way around than what the current mix generates, e.g. lib/kv/file.ex instead of
  kv/lib/file.ex.
* and there are several examples of quotes around atoms, all of the form :"node@host". When I use this form, iex gives me warnings,
  that this is not necessary if the atoms only contain unicode, numbers, etc.

And, lastly, I want to say that this is a great introduction. It was fun to follow along, especially, if one has a bit of erlang background. 

Dieter

### DIFF
--- a/getting-started/mix-otp/agent.markdown
+++ b/getting-started/mix-otp/agent.markdown
@@ -73,7 +73,7 @@ iex>
 
 As you can see, we can modify the agent state in any way we want. Therefore, we most likely don't want to access the Agent API throughout many different places in our code. Instead, we want to encapsulate all Agent-related functionality in a single module, which we will call `KV.Bucket`. Before we implement it, let's write some tests which will outline the API exposed by our module.
 
-Create a file at `test/kv/bucket_test.exs` (remember the `.exs` extension) with the following:
+Create a file at `kv/test/bucket_test.exs` (remember the `.exs` extension) with the following:
 
 ```elixir
 defmodule KV.BucketTest do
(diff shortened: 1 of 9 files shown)


josevalim:
Hi @dischoen! 👋 

> paths for the source files are specified the other way around than what the current mix generates, e.g. lib/kv/file.ex instead of kv/lib/file.ex.

Mix actually generates `kv/lib/kv/file.ex`. We skipped the first `kv` because some complained that to be redundant/confusing. Any suggestions on how to approach this?

> and there are several examples of quotes around atoms, all of the form :"node@host". When I use this form, iex gives me warnings, that this is not necessary if the atoms only contain unicode, numbers, etc.

You are correct! However, the names in the examples contain a `-` and therefore they need quotes. It is probably best to not warn if the atoms contains `@`, to avoid this confusion, so I will change this in Elixir instead. Thank you for the feedback!

dischoen:
Hi again,

on the atoms: my hostname does not contain a dash, so I get this warning. Therefore, in the general case, the quotes are necessary.

on the paths:
I hope I followed the tutorial correctly, and I created a "standalone project" kv first, with 
`mx new kv --module KV`.
This produced the directories:

> kv
>├── lib
>└── test

Later, when umbrella projects are introduced, e.g. with kv_server, this has the following layout:

>kv_umb/
>├── apps
>│   └── my_app
>│       ├── lib
>│       │   └── my_app
>│       └── test
>└── config

There a subdirectory with the name of the application is created within the lib/ directory.
So, for an application inside of an umbrella project this is ok, 
but at least in my case I do not have a subdirectory with the app name inside of the test/ directory.

Anyway, if it is clear for most other users, then I do not have a problem with leaving it as it is.
I mainly took this effort along rewriting, testing and mostly playing with the code in getting-started/.

josevalim:
Generally speaking, Elixir doesn't care about your filenames, so both ways work. But as a convention, a module named KV.Server would be placed at `lib/kv/server.ex`, even if the full path name duplicates it as `apps/kv/lib/kv/server.ex`. :)

josevalim:
Closing this for now as the Paths are correct and the warning has been fixed in Elixir, thanks!

PR #1426, opened 2020-07-25

andrewmarkallen:
Fixed some confusing and incorrect language to be clearer:

"Not only this test cannot run asynchronously, it also requires the expensive setup of stopping and starting the :kv application." ->
"Not only can this test not run asynchronously, but it also requires the expensive setup of stopping and starting the `:kv` application."


josevalim:

PR #1428, opened 2020-07-29
Update recursion.markdown

tisd:
"Mutating is not possible in Elixir" -> "Mutation is not possible in Elixir"

### DIFF
--- a/getting-started/recursion.markdown
+++ b/getting-started/recursion.markdown
@@ -17,7 +17,7 @@ for(i = 0; i < sizeof(array); i++) {
 }
 ```
 
-In the example above, we are mutating both the array and the variable `i`. Mutating is not possible in Elixir. Instead, functional languages rely on recursion: a function is called recursively until a condition is reached that stops the recursive action from continuing. No data is mutated in this process. Consider the example below that prints a string an arbitrary number of times:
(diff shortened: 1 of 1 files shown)


getting-started/recursion.markdown
```diff
@@ -17,7 +17,7 @@ for(i = 0; i < sizeof(array); i++) {
 }
 ```
 
-In the example above, we are mutating both the array and the variable `i`. Mutating is not possible in Elixir. Instead, functional languages rely on recursion: a function is called recursively until a condition is reached that stops the recursive action from continuing. No data is mutated in this process. Consider the example below that prints a string an arbitrary number of times:
+In the example above, we are mutating both the array and the variable `i`. Mutation is not possible in Elixir. Instead, functional languages rely on recursion: a function is called recursively until a condition is reached that stops the recursive action from continuing. No data is mutated in this process. Consider the example below that prints a string an arbitrary number of times:
```
josevalim, line 20:
```suggestion
In the example above, we are mutating both the array and the variable `i`. However, data structures in Elixir are immutable. For this reason, functional languages rely on recursion: a function is called recursively until a condition is reached that stops the recursive action from continuing. No data is mutated in this process. Consider the example below that prints a string an arbitrary number of times:
```

josevalim: Good catch! However, we do have mutations in Elixir in some specific scenarios, so I think it would even be more precise to say our data structures are immutable. :) So I have amended it.  Thanks!

josevalim:

PR #1432, opened 2020-08-19
Modules and Functions: Default Arguments code example fix

maxvoltage:
The code example after line 237 is incorrect.
(Maybe an incomplete update?)

### DIFF
--- a/getting-started/modules-and-functions.markdown
+++ b/getting-started/modules-and-functions.markdown
@@ -235,13 +235,11 @@ If a function with default values has multiple clauses, it is required to create
 
 ```elixir
 defmodule Concat do
-  def join(a, b \\ nil, sep \\ " ")
-
   def join(a, b, _sep) when is_nil(b) do
     a
   end
 
-  def join(a, b, sep) do
+  def join(a, b, sep \\ " ") do
     a <> sep <> b
   end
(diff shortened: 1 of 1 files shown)


eksperimental:
Your code doesn't work because you are not defining join/1.

```elixir
iex(5)> IO.puts Concat.join("Hello")               #=> Hello
** (UndefinedFunctionError) function Concat.join/1 is undefined or private. Did you mean one of:

      * join/2
      * join/3

    Concat.join("Hello")
```

eksperimental:
> The code example after line 237 is incorrect.
> (Maybe an incomplete update?)

what is wrong with the current code?

eksperimental:
Actually the line that you deleted (the function head) is the one the paragraph refers to "If a function with default values has multiple clauses, it is required to create a function head (without an actual body) for declaring defaults".

maxvoltage:
Ah, I am blind! Thank you very much for commenting. I should also have checked my change before opening a PR.
A small suggestion. I think adding a comment next to the function head would be helpful for people who haven't encountered function head before.

`def join(a, b \\ nil, sep \\ " ")  # Function head`

I must have read the phrase you mentioned yet was confused when I saw the function head in the code.

PR #1435, opened 2020-08-25
improve clarity and add missing code

felipelincoln:
The `@impl true` is to keep consistency, and the file path mention is to make the reading less laborious. :grin:
[ETS Guide](https://elixir-lang.org/getting-started/mix-otp/ets.html)


josevalim:

PR #1436, opened 2020-09-01
Added information about mix test starting the full application prior …

Youenn-Bouglouan:
…to running tests

It might be unclear to Elixir newcomers (like myself) that 'mix test' actually starts your whole application before running tests.
This should make it more obvious.

### DIFF
--- a/getting-started/mix-otp/dynamic-supervisor.markdown
+++ b/getting-started/mix-otp/dynamic-supervisor.markdown
@@ -160,7 +160,7 @@ setup do
 end
 ```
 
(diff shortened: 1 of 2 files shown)


josevalim:
Good find @Youenn-Bouglouan! I feel it may be better to talk about this when we talk about starting applications here: https://elixir-lang.org/getting-started/mix-otp/supervisor-and-application.html#starting-applications WDYT? Could you please add a paragraph there? Tks!

Youenn-Bouglouan:
Hi @josevalim sure that makes sense, I added the same information to the section about starting applications.
I feel it makes sense to leave it as a quick reminder on the dynamic supervisor page as well, but I can remove it if you prefer.

Very helpful and pleasant documentation BTW, thank you for putting this together!

josevalim:
I have inlined the notes into the text, thank you!

josevalim:

PR #1439, opened 2020-09-19
Fix typo in distributed-tasks.markdown

rkh:
Replaced "set" with "sent".
I'm also not sure if this is the best way to describe the situation? I think I'd rather call it output rather than input?

> knows all of the *output* should be *sent* back to the original node!

### DIFF
--- a/getting-started/mix-otp/distributed-tasks.markdown
+++ b/getting-started/mix-otp/distributed-tasks.markdown
@@ -73,7 +73,7 @@ iex> Node.spawn_link :"foo@computer-name", fn -> Hello.world end
 hello world
 ```
 
-Elixir spawned a process on another node and returned its pid. The code then executed on the other node where the `Hello.world/0` function exists and invoked that function. Note that the result of "hello world" was printed on the current node `bar` and not on `foo`. In other words, the message to be printed was sent back from `foo` to `bar`. This happens because the process spawned on the other node (`foo`) knows all of the input should be sent back to the original node!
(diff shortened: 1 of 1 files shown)


getting-started/mix-otp/distributed-tasks.markdown
```diff
@@ -73,7 +73,7 @@ iex> Node.spawn_link :"foo@computer-name", fn -> Hello.world end
 hello world
 ```
 
-Elixir spawned a process on another node and returned its pid. The code then executed on the other node where the `Hello.world/0` function exists and invoked that function. Note that the result of "hello world" was printed on the current node `bar` and not on `foo`. In other words, the message to be printed was sent back from `foo` to `bar`. This happens because the process spawned on the other node (`foo`) knows all of the input should be set back to the original node!
+Elixir spawned a process on another node and returned its pid. The code then executed on the other node where the `Hello.world/0` function exists and invoked that function. Note that the result of "hello world" was printed on the current node `bar` and not on `foo`. In other words, the message to be printed was sent back from `foo` to `bar`. This happens because the process spawned on the other node (`foo`) knows all of the input should be sent back to the original node!
```
josevalim, line 76:
```suggestion
Elixir spawned a process on another node and returned its pid. The code then executed on the other node where the `Hello.world/0` function exists and invoked that function. Note that the result of "hello world" was printed on the current node `bar` and not on `foo`. In other words, the message to be printed was sent back from `foo` to `bar`. This happens because the process spawned on the other node (`foo`) knows all of the output should be sent back to the original node!
```

josevalim:
Hi @rkh ! :wave: Long time no see!

You are correct on both counts, merged! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

PR #1445, opened 2020-09-26
getting-started/binaries-strings-and-char-lists#binaries: add a commnt to code for clarification.

jtagcat:
I already forgot, that the default is 8, took a minute.


josevalim:
Thanks @jtagcat but I would prefer to reinforce this information in the text rather than in code comments, if necessary.

jtagcat:
ok :/

PR #1450, opened 2020-10-01
Improve Grammar

ananyasharmawoki:


josevalim:
We prefer those snippets as is, as they are mostly cosmetic. Thanks.

PR #1458, opened 2020-10-10
Update case-cond-and-if.markdown: slightly smoother phrasing

CrownedPhoenix:
Updated to use slightly smoother phrasing.


josevalim:

PR #1460, opened 2020-10-13
Update error output message in binaries page

Coteh:
I'm currently learning Elixir and noticed that the error message I got in 1.11.0 for this part is different than what was shown in the doc. I've updated it with the output I got, slightly modified to remove Elixir version number from output.


josevalim:

PR #1462, opened 2020-10-24
Improve sidebar in the getting-started section (a little bit)

azlkiniue:
I'm learning elixir now, but I feel lost when looking at the guides because the links looks all the same. So here I added a little control flow to disable the link of current page.


josevalim:

PR #1464, opened 2020-10-27
Getting started/Basic operators: Comment above `!` operator should be `not`

tiosgz:
Because `&&` has `and` and `||` has `or`


josevalim:

PR #1467, opened 2020-11-04
Update agent.markdown

vukanac:
Update with the real output in
Erlang/OTP 23 [erts-11.1.1] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [hipe] [dtrace]
Interactive Elixir (1.11.2)


josevalim:
Hi @vukanac! Unfortunately this change is not correct, can you please double check your commands? thank you!

vukanac:
Hi @josevalim, you are totally right. Definitely my mistake. Accidentally I started with `{}`.
Thanks! Closing.

PR #1469, opened 2020-11-19
Add example of @callback with @doc tag

vnbrs:
This PR adds a `@doc` tag along with the `@callback` tag. I think this might help to show where the doc usually is when using behaviours.


josevalim:

PR #1472, opened 2020-11-23
Update case-cond-and-if.markdown

akrk1986:
seems like a typo to me


josevalim:

PR #1473, opened 2020-12-03
fix(getting-started): add a note about scripting files execution

psylone:
Noticed that we can execute both `.ex` and `.exs` files with `elixir` command and no bytecode file will be created. Probably we need a note about that.

### DIFF
--- a/getting-started/modules-and-functions.markdown
+++ b/getting-started/modules-and-functions.markdown
@@ -62,7 +62,7 @@ When working on actual projects, the build tool called `mix` will be responsible
 
 ## Scripted mode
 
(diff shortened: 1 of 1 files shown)


josevalim:
Thank you @psylone! This is correct but since this is a getting started guide and not a complete reference, I am worried about mentioning `mix` early on and confusing users. Thoughts?

psylone:
On the one side yes, although `mix` is mentioned in the last paragraph right before `Scripted mode` section (we can also use a link to hexdocs for `mix`).

Also (as for me) it's a little bit unclear when you experiment with `.ex` and `.exs` files and get the same result (no `.beam` file created). I remember it was one of the moment for research when I started learn Elixir ))

josevalim:
Thank you! I have streamlined the changes based on your comments and suggestions.

josevalim:

psylone:
Thanks!

PR #1478, opened 2020-12-22
Minor edit: use matching module to query dynamic supervisor

Odaeus:
`Supervisor` and `DynamicSupervisor` both have `count_children` and both appear to work identically in this situation. However, `KV.BucketSupervisor` is a `DynamicSupervisor` so we should probably use functions from the matching module.

While reading the docs I stumbled across this and had to go back up a bit to double-check it was still referring to the dynamic supervisor.


josevalim:

PR #1485, opened 2021-01-26
Fix colon position.

taku333:
The colon was in the wrong place, so I fixed it.


josevalim:
It can be a bit confusing but it is correct. only and except are atoms, so the colon is leading, but they are trailing when used as keywords. In any case, thanks for the PR!

PR #1486, opened 2021-01-29
Update io-and-the-file-system.markdown

nics:
To be consistent char_data should be spelled chardata?


josevalim:

PR #1491, opened 2021-02-11
fix(otp): remove extra assert

psylone:
It looks like we don't need `assert` when lookup created registry.


josevalim:
We don't need it but it doesn't hurt either and it often has better error messages in case it does fail. So I prefer to keep it. :) Thank you!

PR #1494, opened 2021-02-17
Highlight mix test output on the homepage

wojtekmach:
This might be trying a little bit too hard but I though I'd send it anyway! :)

Before:

image

After:

image


josevalim:

PR #1500, opened 2021-03-11
Add more reasons to use a version manager

axelson:
From what I have seen, version managers are widely recommended in the
Elixir community, so it feels appropriate to add another reason to use
them.


josevalim:

PR #1501, opened 2021-03-15
Nobody uses feedburner anymore. We should use the original feed

derek-zhou:
The feed url on the website is confusing, and do not work in many places. The original atom feed is good enough; there is no point to use the obsolete feedburner service.


josevalim:

PR #1502, opened 2021-03-24
Remove confusing `use` statement

DukeFruitarian:
I think we should not `use Agent` here, since we use `Agent`'s functions  directly with module name everywhere. It may be confusing to `use` it inside `KV.Bucket` module.

### DIFF
--- a/getting-started/mix-otp/agent.markdown
+++ b/getting-started/mix-otp/agent.markdown
@@ -103,8 +103,6 @@ In order to fix the failing test, let's create a file at `lib/kv/bucket.ex` with
 
 ```elixir
 defmodule KV.Bucket do
-  use Agent
-
   @doc """
   Starts a new bucket.
   """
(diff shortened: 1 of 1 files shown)


josevalim:
`use Agent` does not import the agent functions. It does, however, define a `child_spec` function with the default behaviours, which is why we use it. :) Perhaps that should be made clearer instead? Thanks for the PR!

DukeFruitarian:
> `use Agent` does not import the agent functions. It does, however, define a `child_spec` function with the default behaviours, which is why we use it. :) Perhaps that should be made clearer instead? Thanks for the PR!

oh, yeah, it wasn't obvious for me, thanks.

DukeFruitarian:
But anyway i can't see where and for what we use `child_spec` in 2-nd chapter.  Please write it more explicit if we do :-) And may be add it later if it's used in later chapters.

josevalim:
Yes, it won't come in use until we add supervisors. I will add the notes.

DukeFruitarian:
> Yes, it won't come in use until we add supervisors. I will add the notes.
Not everything is clear. Thanks. So i close pr than.

PR #1503, opened 2021-03-25
Update module attributes section

japhib:
Updating module attributes section with note about how attributes defined with functions are processed at compile-time, not runtime.

### DIFF
--- a/getting-started/module-attributes.markdown
+++ b/getting-started/module-attributes.markdown
@@ -85,8 +85,6 @@ defmodule MyServer do
 end
 ```
 
-> Note: Unlike Erlang, user defined attributes are not stored in the module by default. The value exists only during compilation time. A developer can configure an attribute to behave closer to Erlang by calling [`Module.register_attribute/3`](https://hexdocs.pm/elixir/Module.html#register_attribute/3).
-
 Trying to access an attribute that was not defined will print a warning:
 
 ```elixir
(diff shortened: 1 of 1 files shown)


josevalim:
Hi @japhib. Thanks for the changes but at this point of the guide we haven't introduced Mix at all. Perhaps, if we use module attributes in the Mix guide, we can add a note there?

japhib:
Thanks for the quick response! I've updated the changes so the example is a bit more concise, and doesn't reference Mix at all. Would this be better?

getting-started/module-attributes.markdown
```diff
@@ -135,6 +135,16 @@ end
...
 
+Note that the function will be called at compilation time and its *return value*, not the function call itself, is what will be substituted in for the attribute. So, if `URI.parse("https://example.com")` returns a struct of `%URI{}`, then the above example will compile to something like this:
+
+```elixir
+defmodule MyApp.Status do
+  def status(email), do: SomeHttpClient.get(%URI{})
+end
+```
+
+This can be useful for pre-computing constant values, but it can also cause problems if you're expecting the function to be called at runtime. So if you are e.g. reading a value from a database or the environment inside an attribute, be aware that it will read that value only at compilation time.
```
japhib, line 146: I'm open to suggestions on how to word this -- obviously it's a useful feature, and I don't want to imply that it causes problems. The problem is just if you don't understand how it works. 

Maybe instead of `it can cause problems`, should I change it to `it can be a source of confusion` or something? Or if you don't care, then we can just leave it I guess

getting-started/module-attributes.markdown
```diff
@@ -135,6 +135,16 @@ end
 
 Be careful, however: *functions defined in the same module as the attribute itself cannot be called* because they have not yet been compiled when the attribute is being defined.
 
+Note that the function will be called at compilation time and its *return value*, not the function call itself, is what will be substituted in for the attribute. So, if `URI.parse("https://example.com")` returns a struct of `%URI{}`, then the above example will compile to something like this:
+
+```elixir
+defmodule MyApp.Status do
+  def status(email), do: SomeHttpClient.get(%URI{})
```
josevalim, line 142: Can we replace this by the literal struct result? :) I think it will make it clearer.

getting-started/module-attributes.markdown
josevalim: I think this wording is good. I can tweak it if necessary right before merging. :)

getting-started/module-attributes.markdown
japhib: Sure!

japhib: The only reason I didn't include it originally was that it's kind of long:

```
iex(1)> URI.parse("https://example.com")
%URI{
  authority: "example.com",
  fragment: nil,
  host: "example.com",
  path: nil,
  port: 443,
  query: nil,
  scheme: "https",
  userinfo: nil
}
```

So, just thinking of options here:

1. Include the whole long thing
2. Shorten it by removing the `nil` fields
3. Change the example to something short like `DateTime.from_unix!`

Which would you prefer? I'm leaning towards option 3, personally

josevalim: I think 2 works great. Showing a big example is good, because it shows precisely how you can end-up inserting tons of data!

japhib: Great, I'll make that change 👍

japhib: I've pushed that change

josevalim:
Beautiful, I slightly restructured it and pushed another pitfall! 🔥

josevalim:

japhib:
Thanks! I like this flow of the article better now, with your changes. Also, this is my first open-source Elixir contribution 😄

josevalim:
Congratulations and thank you! 🎉

PR #1507, opened 2021-03-31
Remove `usually`

DukeFruitarian:
1. Remove second "usually" word.

2. I think it's worth to note in guide that after we change `bar` release to use only `kv` app and add cookies to the both releases, we MUST reassemble `foo` one too. Otherwise cookies between nodes won't match, since we still have working release where `foo` uses autogenerated value.


josevalim:

PR #1513, opened 2021-04-18
Add missing piece of config info to instructions

bryszard:
To be able to use the `distributed` tag for the exclusion afaik we need the same config in `test_helper.exs` for `:kv_server` as we do for `:kv`. This instruction is implicit and I think it would be better to have it stated explicitly.


josevalim:

PR #1514, opened 2021-04-18
Add more info to the Assembling multiple releases

bryszard:
At this point of tutorial the release with our application is working, but only partially. This information is implicit and I'm making it explicit, by making a request with valid input and describing it.


josevalim:

PR #1515, opened 2021-04-18
Getting Started: Reassemble after cookie change

ryanbrainard:
In the Getting Started: Mix & OTP guide, the `cookie` changes in `foo`, but it does not instruct the learner to reassemble `foo`, so it results in an `Invalid challenge reply.`


josevalim:

PR #1516, opened 2021-04-26
Removing font weight property for better readability.

chin-chillin:
Current font weight setting "0.95em/1.692307em" produce thinner fonts than the default. Default font weight setting is more comfortable to the eyes for longer reading sessions. I found the default font weight work best with dark reader addon as well. Here's how the website renders with current "0.95em/1.692307em" setting,

no-change

Removing the font weight and leaving it to the browser,

font-weight-removed

I do understand this boils down to personal preference and anyone can change this through a browser extension like Stylus. In that case, feel free to ignore this.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -33,7 +33,7 @@ html {
 }
 body {
   margin: 0;
-  font: 0.95em/1.692307em 'Bitter', Georgia, 'Times New Roman', Times, serif;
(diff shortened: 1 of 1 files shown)


josevalim:
Thanks, I gave it a try and I was not sure it is best, so I prefer to go with no change for now. :)

PR #1517, opened 2021-05-07
Remove @before_compile from popular module attrs

v0idpwn:
It isn't popular, as implied by the text, and it doesn't seem to me that beginners would benefit from this knowledge.

Add `@spec` instead. While there is already a section that cover type specifications (the same one that covers behaviours),
I think it is better to first mention them here instead of waiting as long as the guide does currently.

If you want to inform the reader that module attributes can be used to add compilation hooks, I can add a small section about
it in the end of the page.


josevalim:

Issue #1521, opened 2021-05-17
Mix Guide


ThiagosLima:
Hi, I was reading the Introduction to Mix section on the Getting Started Guide and get a little confused. It says:

> Let’s create our first project by invoking `mix new` from the command line. **We’ll pass the project name as the argument** (`kv`, in this case), and tell Mix that our main module should be the all-uppercase `KV`, instead of the default, which would have been `Kv`:

But the `mix help new` says:

> Creates a new Elixir project. **It expects the path of the project as argument.**
>    `mix new PATH [--app APP] [--module MODULE] [--sup] [--umbrella]`
>A project at the given PATH will be created. The application name and module
name will be retrieved from the path, unless `--module` or `--app` is given.

Maybe if the guide says something like this would be less confusing.

> Let’s create our first project by invoking `mix new` from the command line. **We’ll pass the project path as the argument (`kv`, in this case). By default the application name and module name will be retrieved from the path**. So we tell Mix that our main module should be the all-uppercase `KV`, instead of the default, which would have been `Kv`:

josevalim:
Great idea! Can you please submit a PR?

ThiagosLima:
Nice! I will do it.

PR #1522, opened 2021-05-17
Update mix new description (#1521)

ThiagosLima:
This is the updated description of `mix new` accordingly to the issue #1521.


whatyouhide:
Thank you @ThiagosLima! 💟

PR #1530, opened 2021-06-08
Update module-attributes.markdown

angrycandy:
Fixed typo: it should be either _as_ or _for_ but not both. Use _as_ because it matches the rest of the section.


josevalim:

PR #1532, opened 2021-06-09
Update the example for variables visibility in the do-block of `try`

nicnilov:
The `iex> what_happened` line is unnecessary. The `** (RuntimeError) undefined function: what_happened/0` will be caused by an attempt to access the variable within the `rescue` block. This line will attempt to do the same again, which will only demonstrate that the variables defined inside the `try` block do not leak, as was already shown by the preceding example.


josevalim:

josevalim:

PR #1538, opened 2021-06-22
Update introduction-to-mix.markdown

JRhodes95:
Fixed the tiniest typo where a word in the testing sentence had crept into a code snippet.


josevalim:
The current approach is intentional, after all, there is a `use ExUnit.Case` in the code, but I am glad to accept this change if you still think it is not clear. :) thanks!

JRhodes95:
Ah I see what you mean, apologies! Newbie learning this stuff for the first time and thought I'd just fix it as I saw it. Closing PR as I understand the meaning now.

PR #1540, opened 2021-07-05
Update pattern-matching.markdown

trkoch:
It's easy to confuse `{` and `[`. It happened to me and the code example at first didn't make sense. Clarify by mentioning types of example.


josevalim:

PR #1543, opened 2021-07-20
Handle n = 0 for Recursion.print_multiple_times/2

KNejad:
The previous implementation would print "Hello!" even if `n` was set to 0 or a negative number.

With this implementation, when 0 is passed, it will not print anything and when a negative number is passed we will get a `FunctionClauseError` which I believe is more appropriate.

I understand this is just meant for illustration purposes of how to perform "loops" in Elixir, but figured this doesn't complicate it much more and produces better results.

If this is seen as over-complicating the example, I'm happy to close the PR though :)

### DIFF
--- a/getting-started/recursion.markdown
+++ b/getting-started/recursion.markdown
@@ -18,11 +18,10 @@ In the example above, we are mutating both the array and the variable `i`. Howev
 
 ```elixir
 defmodule Recursion do
-  def print_multiple_times(msg, n) when n <= 1 do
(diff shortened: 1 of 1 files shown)


josevalim:
Good call, this is definitely more idiomatic. I also flipped the clauses around and added a `:ok` return.  :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

josevalim:

PR #1545, opened 2021-07-24
Update to the output of ExUnit 1.12

keshihoriuchi:
The output of the current version of ExUnit is different from here.

We need `assert` because this output should come from https://github.com/elixir-lang/elixir-lang.github.com/blob/01af1d9f2a14956371361ef55c663ed415e565bc/getting-started/mix-otp/genserver.markdown?plain=1#L245 .


josevalim:

PR #1548, opened 2021-07-30
Fix mini documentary link

soto-ylopo:
The link on the page seems to be a dead one. I believe this is the the updated link https://cult.honeypot.io/originals/elixir-the-documentary


josevalim:

PR #1551, opened 2021-08-08
Add SpawnFest for 2021

starbelly:
Add SpawnFest hack-a-thon for 2021

Not sure what this will look like, tried to build locally but lots of dependency issues.

Edit: 

If the logo is way too big (i.e., there's no css rules to normalize it) let me know and I can put up a small logo on the site. If so, can you suggest a size?


josevalim:
Done, thanks!

Issue #1553, opened 2021-08-14
Example about Unicode characters mixes up bytes and characters


dcz-self:
The text mentions Unicode characters (rather unambiguously code points), but the example counts graphemes and bytes, and not characters:

> In other words, what we would expect to be a single character, such as `é` or `ł`, can in practice be multiple characters, each represented by potentially multiple bytes. Consider the following:
>
> ```elixir
> iex> string = "hełło"
> "hełło"
> iex> String.length(string)
> 5
> iex> byte_size(string)
> 7
```

A better string would be "héllo", which is "e" + "U+0301 COMBINING ACUTE ACCENT" (there's no standalone stroke of the "ł" that I could find, making the "can be multiple characters" assertion doubtful). It would result in:
- 5 graphemes
- 6 code points
- 7 bytes

I think what's missing from the tutorial is a way to count code points.

josevalim:
Excellent catch. Let's remove `ł` from the list and use `héllo` instead. Probably say something like this:

> In other words, what we would expect to be a single character, such as é or ł, can in practice be multiple codepoints, each represented by potentially multiple bytes. Consider the following:

What do you think? Would you like to send a PR? There are a couple ways you could count codepoints but, to be honest, I wouldn't focus on that, because it is the least used of them.

dcz-self:
There's still the trouble with "ł", where I couldn't find a way to represent it with multiple codepoints. Regarding code points, I have no idea how to count them. I think they are important for interoperability, because graphemes have their own baggage: popular languages like Rust, Python don't have a simple way to calculate them, and on top of that they can fall apart into multiple ones. E.g. emoji + ZWJ + modifier may get collapsed into a single grapheme or not, depending on the available font.

I propose:

> In other words, what we would expect to be a single character, such as é, can in practice be multiple codepoints (in this case, e and an acute accent), each represented by potentially multiple bytes. Consider the following:

If you show me how to count code points, I'm going to submit the result.

josevalim:
I meant to remove ł from the text but I forgot. :D

Your suggestion is perfect. `length(String.to_charlist(...))` can retrieve the codepoints.

dcz-self:
Done in https://github.com/elixir-lang/elixir-lang.github.com/pull/1554

josevalim:
Cool, thank you. Closing in favor of the PR.

PR #1554, opened 2021-08-15
Correct code point explanation

dcz-self:
The example now has a grapheme made of multiple code points. The "ł" example was removed, as it's not clear how to compose it out of 2 code points.

https://github.com/elixir-lang/elixir-lang.github.com/issues/1553

### DIFF
--- a/getting-started/binaries-strings-and-char-lists.markdown
+++ b/getting-started/binaries-strings-and-char-lists.markdown
@@ -62,7 +62,7 @@ iex> byte_size(string)
 7
 ```
 
-`String.length/1` counts graphemes, but `byte_size/1` reveals the number of underlying raw bytes needed to store the string when using UTF-8 encoding. UTF-8 requires one byte to represent the characters `h`, `e`, and `o`, but two bytes to represent `ł`.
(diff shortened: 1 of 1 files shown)


getting-started/binaries-strings-and-char-lists.markdown
```diff
@@ -49,13 +49,15 @@ Elixir uses UTF-8 to encode its strings, which means that code points are encode
...
-iex> string = "hełło"
-"hełło"
+iex> string = "héllo"
+"héllo"
 iex> String.length(string)
 5
+iex> length(String.to_charlist(string))
+6
 iex> byte_size(string)
 7
```
josevalim, line 62:
```suggestion
iex> string = "héllo"
"héllo"
# Grapheme count
iex> String.length(string)
5
# Codepoint count
iex> length(String.to_charlist(string))
6
# Byte count
iex> byte_size(string)
7
```

josevalim: I am thinking adding some comments can help make the code clearer. WDYT?

josevalim: Alternatively we can update the paragraph below. I think that will actually be clearer. Let me try that.

josevalim:

PR #1556, opened 2021-08-16
Add a page on optional syntax

josevalim:


getting-started/optional-syntax.markdown
```diff
@@ -0,0 +1,89 @@
...
+...> else
+...>   :that
+...> end
+
+# keyword lists
+iex> if true, do: :this, else: :that
+:this
+```
+
+Those affordances, which we call here "optional syntax", allow the language syntax core to be small, without sacrificing the readability and expressivess of your code. This is what allows us to write:
```
fertapric, line 29:
```suggestion
Those affordances, which we call here "optional syntax", allow the language syntax core to be small, without sacrificing the readability and expressiveness of your code. This is what allows us to write:
```

getting-started/optional-syntax.markdown
```diff
@@ -0,0 +1,89 @@
...
+
+4. Finally, parentheses are optional, but let's add them:
+
+   ```elixir
+   if(variable?, [{:do, Call.this()}, {:else, Call.that()}])
+   ```
+
+That's it! Those four rules outline the optional syntax of the majority of the code we have written so far. Whenever you have any questions, this quick walk-through has you covered.
+
+On their day to day, Elixir developers use the [`mix format`](https://hexdocs.pm/mix/Mix.Tasks.Format.html) task, which formats your codebase according to a well-defined set of rules defined by the Elixir team and the community. For instance, `mix format` will always add parentheses to function calls unless explicitly configured to not do so. This ensures all codebases in your company and in the community follow the same standards.
```
fertapric, line 89:
```suggestion
On their day to day, Elixir developers use the [`mix format`](https://hexdocs.pm/mix/Mix.Tasks.Format.html) task to format their codebase according to a well-defined set of rules defined by the Elixir team and the community. For instance, `mix format` will always add parentheses to function calls unless explicitly configured to not do so. This ensures all codebases in your company and in the community follow the same standards.
```

fertapric (approved):
😍

getting-started/optional-syntax.markdown
```diff
@@ -0,0 +1,89 @@
+---
+section: getting-started
+layout: getting-started
+title: Optional syntax sheet
+---
+
+Over the several chapters, we have learned that Elixir syntax has a few affordances where developers may omit some delimiters to make the code more readable. For example, we have learned that parentheses are optional:
```
whatyouhide, line 7: Does this read simpler?

```suggestion
Over several chapters we learned that the Elixir syntax has a few instances where developers may omit some delimiters to make code more readable. For example, we have learned that parentheses are optional:
```

getting-started/optional-syntax.markdown
```diff
@@ -0,0 +1,89 @@
...
+
+```elixir
+iex> length([1, 2, 3]) == length [1, 2, 3]
+true
+```
+
+and that do-end blocks are equivalent to keyword lists:
```
whatyouhide, line 14: Maybe?

```suggestion
and that `do`/`end` blocks are equivalent to keyword lists:
```

getting-started/optional-syntax.markdown
whatyouhide: I see that you use "affordances" again but to be honest I am struggling a bit to understand what exactly that means, which is why I suggested a "simpler" word above 😄

getting-started/optional-syntax.markdown
whatyouhide:
```suggestion
In their day to day, Elixir developers use the [`mix format`](https://hexdocs.pm/mix/Mix.Tasks.Format.html) task, which formats your codebase according to a well-defined set of rules defined by the Elixir team and the community. For instance, `mix format` will always add parentheses to function calls unless explicitly configured to not do so. This ensures all codebases in your company and in the community follow the same standards.
```

getting-started/optional-syntax.markdown
```diff
@@ -0,0 +1,89 @@
...
+
+instead of:
+
+```elixir
+defmodule(Math, [
+  {:do, def(add(a, b), [{:do, a + b}])}
+])
+```
+
+In this brief chapter, we will review the 4 affordances provided by the language, using a short snippet as playground.
```
whatyouhide, line 47:
```suggestion
In this brief chapter, we will review the four affordances provided by the language, using a short snippet as playground.
```

You use words below, so I think let's use words here too :)

whatyouhide (approved):
This is awesome, definitely a very welcome addition 💯

josevalim:

PR #1558, opened 2021-08-27
Minor improvements to the optional syntax page

lexmag:


getting-started/optional-syntax.markdown
```diff
@@ -4,14 +4,14 @@ layout: getting-started
...
-Throughout this guide, we learned that the Elixir syntax allow developers to omit delimiters in a few occasions to make code more readable. For example, we learned that parentheses are optional:
+In this guide so far, we learned that the Elixir syntax allows developers to omit delimiters in a few occasions to make code more readable. For example, we learned that parentheses are optional:
 
 ```elixir
 iex> length([1, 2, 3]) == length [1, 2, 3]
 true
 ```
 
-and that `do`/`end` blocks are equivalent to keyword lists:
+and that `do`-`end` blocks are equivalent to keyword lists:
```
josevalim, line 14:
```suggestion
and that `do`/`end` blocks are equivalent to keyword lists:
```

getting-started/optional-syntax.markdown
```diff
@@ -60,7 +60,7 @@ end
 
 Now let's remove the conveniences one by one:
 
-1. do-end blocks are equivalent to keywords:
+1. `do`-`end` blocks are equivalent to keywords:
```
josevalim, line 63:
```suggestion
1. `do`/`end` blocks are equivalent to keywords:
```

josevalim:
We use `do`/`end` in Elixir, so I reverted to that, but I don't oppose moving to `do`-`end` as long as we are consistent. :)

josevalim:

lexmag:
> We use do/end in Elixir, so I reverted to that, but I don't oppose moving to do-end as long as we are consistent. :)

Indeed, we also have `do/end` occurrences. I'll standardize everything on `do`-`end` then (it reads like a continuous block, while `/` can be perceived as a choice; of course, the exact context makes it less problematic but still).

Issue #1559, opened 2021-09-02
Update dynamic-supervisor.markdown


mattjg908:
In the elixir-lang.org docs [here](https://elixir-lang.org/getting-started/mix-otp/dynamic-supervisor.html#supervision-trees), the KV.Supervisor's `init` is updated to be:
```
  def init(:ok) do
    children = [
      {DynamicSupervisor, name: KV.BucketSupervisor, strategy: :one_for_one},
      {KV.Registry, name: KV.Registry}
    ]

    Supervisor.init(children, strategy: :one_for_all)
  end
```

[The text then says](https://elixir-lang.org/getting-started/mix-otp/dynamic-supervisor.html#shared-state-in-tests):
> Since we have now changed our registry to use KV.BucketSupervisor, which is registered globally

I _believe_ the text above uses the term "globally" a little loosely because `init` does not register `KV.BucketSupervisor` globally, it registers it locally.

I think the intention was to say something more along the lines of "Since we have now changed our registry to use KV.BucketSupervisor, which is registered locally and can now be referenced by the name `KV.BucketSupervisor`,.." or something along those lines.

Perhaps the text could/should be updated to be:
> Since we have now changed our registry to use KV.BucketSupervisor, our tests are now relying on this shared supervisor even though each test has its own registry. The question is: should we?

Any thoughts?

josevalim:
Yes, good catch! Please send  a PR!

mattjg908:
> Yes, good catch! Please send a PR!

Great, I will do so1 :)

PR #1560, opened 2021-09-02
Update dynamic-supervisor.markdown

mattjg908:
## Reason for change
See [issue 1559](https://github.com/elixir-lang/elixir-lang.github.com/issues/1559)

## Summary of changes
This PR changes the text from:
> Since we have **now** changed our registry to use KV.BucketSupervisor, **which is registered globally,** our tests are now relying on this shared supervisor even though each test has its own registry.

To:
> Since we have changed our registry to use KV.BucketSupervisor, our tests are now relying on this shared supervisor even though each test has its own registry.

#### Before
Screen Shot 2021-09-02 at 11 43 34 PM

#### After
Screen Shot 2021-09-02 at 11 45 19 PM

### DIFF
--- a/getting-started/mix-otp/dynamic-supervisor.markdown
+++ b/getting-started/mix-otp/dynamic-supervisor.markdown
@@ -158,7 +158,7 @@ setup do
 end
 ```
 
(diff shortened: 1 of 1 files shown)


josevalim:

PR #1563, opened 2021-09-17
Fixed incorrect file name

joshua-bouv:
Incredibly minor but tutorial says to add '@tag :distributed' in a test located in `test_helper.exs`, when it actually is in `kv_server_test.exs`.


joshua-bouv:
My mistake - didn't read it properly!

PR #1566, opened 2021-10-01
Explicitly state what the capture operator is

djfpaagman:
The capture operator is introduced here, but it's not explicitly stated what that operator is. I was reading through this part of the getting started guide and was a bit confused as I had to understand the new concept (capture operator) which is already used in the code snippet just above, I think this makes it more clear that it's the same thing you just saw in the code snippet.


josevalim:

PR #1569, opened 2021-10-11
Add an log to avoid typo in another handle_info

Ma233:
To prevent someone makes typo in `def handle_info({:DOWN, ref, :process, _pid, _reason}, {names, refs}) do`, such as `:proccess` for `:process`.
It might be difficult for a new programmer of elixir to find out why the bucket is not deleted properly.

### DIFF
--- a/getting-started/mix-otp/genserver.markdown
+++ b/getting-started/mix-otp/genserver.markdown
@@ -305,6 +305,7 @@ end
 
 @impl true
 def handle_info(_msg, state) do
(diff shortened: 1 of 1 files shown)


josevalim:

Ma233:
@josevalim Hi, It looked strange in the docs now.
image

As a beginner, I'm not sure if this is what you want?
```elixir
  @impl true
  def handle_info(msg, state) do
    require Logger
    Logger.debug("Unexpected message in KV.Registry: #{inspect(msg)}")
    {:noreply, state}
  end
```

PR #1576, opened 2021-11-05
Update recursion.markdown

colmben:
I'm assuming a typo here, it  doesn't work in iex anyways?

### DIFF
--- a/getting-started/recursion.markdown
+++ b/getting-started/recursion.markdown
@@ -133,7 +133,7 @@ iex> Enum.map([1, 2, 3], fn(x) -> x * 2 end)
 Or, using the capture syntax:
 
 ```elixir
-iex> Enum.reduce([1, 2, 3], 0, &(&1 + &2)
+iex> Enum.reduce([1, 2, 3], 0, &(&1 + &2))
 6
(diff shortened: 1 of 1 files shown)


josevalim:
The current example should totally work. I just tried it on my IEx. Can you please show what have you tried and which error you get?

colmben:
Hi Jose!

I was just just reading ahead on a chromebook and so used the online iex at https://try-elixir.herokuapp.com/ which fails on that line. I now see that it does indeed work on a locally installed iex, The syntax totally threw me I have to say, why would you capture a kernel function like that instead of just clearly showing two arguments adding together? It seems to be a case of brevity over clarity, unless that form is common in Elixir? In which case it might be worth mentioning what it is doing in the text? Note I wasn't the only one perplexed by the syntax, when you google "&+/2" the first (and only Elixir relevant) hit is https://stackoverflow.com/questions/39156130/what-is-2-in-elixir/39156167

josevalim:
I see! Someone on reddit was also asking earlier this week what `&+/2` means, so I sent a commit to introduce the concept earlier. Thank you for bringing it up!

PR #1579, opened 2021-11-20
behaviour-protocol

achalagarwal:
Behaviours and Protocols both seem to work like Interfaces (in OOP based languages), so perhaps it might make sense to mention the differentiation aspect between the two.

Also, IIUC, behavior's seem more like an Interface and Protocols seem like Subclassing (because there can be a default implementation there).


josevalim:

PR #1581, opened 2021-12-10
Fix the missing link to the guides page

kentaro:
A path I'd like to fix points to a seemingly obsolete guides page. I'll fix it along with the path that is linked from "GUIDES" in the header menu.

image


josevalim:

PR #1583, opened 2021-12-13
Update Erlang download link

kevinschweikert:
This fixes just the URL of the erlang download link. Otherwise you would see a "Page not found" message


josevalim:

PR #1594, opened 2022-01-29
Use single line if macro

bmarkons:
Given that `IF` is written in a single line maybe it would be more clear if we represented `IF` macro with a corresponding single-line form.

Without this change you get an error in the iex when entering:

```
quote do: if true do :this else :that end
```


josevalim:
Thank you for the PR. Unfortunately I believe the current version is correct. We are asking for you to check how the keyword and is equivalent to the do-block version. For that to work though, you need to put parens around it or use the do-block version for `quote` too.

bmarkons:
Oh, I see. Sorry for the confusion and thank you for your response! Enjoy your weekend 🍹

PR #1599, opened 2022-03-12
fix: string example

nothiel:
String.to_charlist on elixir returns the charlist of an string, which in
this case is [104, 233, 108, 108, 111], so i fixed this example by
putting the correct value

### DIFF
--- a/getting-started/binaries-strings-and-char-lists.markdown
+++ b/getting-started/binaries-strings-and-char-lists.markdown
@@ -57,12 +57,12 @@ iex> string = "héllo"
 iex> String.length(string)
 5
 iex> length(String.to_charlist(string))
-6
+5
 iex> byte_size(string)
-7
+6
 ```
 
-`String.length/1` counts graphemes and returned 5. To count the number of code points, we can use `String.to_charlist/1` to convert a string to a list of codepoints, and then we get its length, which returned 6. Finally, `byte_size/1` reveals the number of underlying raw bytes needed to store the string when using UTF-8 encoding. UTF-8 requires one byte to represent the characters `h`, `e`, `l`, and `o`, but two bytes to represent the acute accent, adding to 7.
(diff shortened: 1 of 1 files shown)


getting-started/binaries-strings-and-char-lists.markdown
```diff
@@ -57,12 +57,12 @@ iex> string = "héllo"
 iex> String.length(string)
 5
 iex> length(String.to_charlist(string))
-6
+5
```
wojtekmach, line 60: > String.to_charlist on elixir returns the charlist of an string, which in this case is `[104, 233, 108, 108, 111]`

could you double-check? For me it is:

```elixir
iex(1)> string = "héllo"
"héllo"
iex(2)> String.length(string)
5
iex(3)> length(String.to_charlist(string))
6
iex(4)> byte_size(string)
7
```

which is what was there before your patch.

Also:

```elixir
iex> String.to_charlist(string)
[104, 101, 769, 108, 108, 111]
```

nothiel: hi there! first of all, i'm using iex and my elixir version is
```
Erlang/OTP 24 [erts-12.2.1] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit]

Elixir 1.13.0 (compiled with Erlang/OTP 24)
```


```elixir
iex(26)> string = "héllo"
"héllo"
iex(27)> String.length(string)
5
iex(28)> length(String.to_charlist(string))
5
iex(29)> byte_size(string)
6
iex(30)> 
```

```elixir
iex(31)> String.to_charlist(string)
[104, 233, 108, 108, 111]
```

nothiel: Tried to run on a docker container to see if this changes something, but it still gives the same result that i passed to you guys...

image

josevalim:

PR #1603, opened 2022-04-13
git unauthenticated protocol is no longer supported

aravindanck:
When I ran mix deps.get with a git repo as my dependency, it threw me the following error. Looks like git unauthenticated protocol is no longer supported for security reasons and https protocol is the way to go. 

mix deps.get
* Getting broadway_kafka (git://github.com/dashbitco/broadway_kafka.git)
fatal: remote error: 
  The unauthenticated git protocol on port 9418 is no longer supported.
Please see https://github.blog/2021-09-01-improving-git-protocol-security-github/


josevalim:

PR #1604, opened 2022-04-21
Add dark logo to use in Elixir README

simonprev:
Since the images from https://github.com/elixir-lang/elixir are hosted in this repo, we need to add a new image to support GitHub dark mode. The result will look like this:

image

**Before**

image


josevalim:

PR #1605, opened 2022-05-03
remove the "apt-key is deprecated" warning

RooSoft:
Trying to install elixir on a Raspberry pi using the [official instructions](https://elixir-lang.org/install.html#raspberry-pi) leads to an error.

The `sudo apt-key add erlang_solutions.asc` command issues this message: 

```
Warning: apt-key is deprecated. Manage keyring files in trusted.gpg.d instead (see apt-key(8)).
```

It's just a warning, but things go wrong starting from that point...

It seems that this is the new way to import the key

```
cat erlang_solutions.asc | gpg --dearmor > erlang_solutions.gpg
sudo install -o root -g root -m 644 erlang_solutions.gpg /etc/apt/trusted.gpg.d/
```

And from that point on, it installs successfully.

### DIFF
--- a/install.markdown
+++ b/install.markdown
@@ -102,7 +102,8 @@ If necessary, replace "buster" with the name of your Raspbian release.
   * Get Erlang key and add it to the keychain:
     * Run: `echo "deb https://packages.erlang-solutions.com/debian buster contrib" | sudo tee /etc/apt/sources.list.d/erlang-solutions.list`
     * Run: `wget https://packages.erlang-solutions.com/debian/erlang_solutions.asc`
-    * Run: `sudo apt-key add erlang_solutions.asc`
+    * Run: `cat erlang_solutions.asc | gpg --dearmor > erlang_solutions.gpg`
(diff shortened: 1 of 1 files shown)


RooSoft:
More on that...

From the repository, `apt install elixir` installs these versions by default

```
Erlang/OTP 23 [erts-11.1.8] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1]

Elixir 1.10.3 (compiled with Erlang/OTP 22)
```

I would have expected at least Elixir 1.13.x and OTP 24...

Is there something I could do to help push those versions up a notch?

josevalim:
We don’t maintain those packages, but you can reach out to the maintainers. :) thanks for the PR!

josevalim:

RooSoft:
> We don’t maintain those packages, but you can reach out to the maintainers. :) thanks for the PR!

Sent a message to Erlang Solutions on their website...

Hoping they'll answer in a timely fashion, this outdated package isn't a good look.

PR #1616, opened 2022-07-04
Use port 4321 in telnet command (Task and gen_tcp guide)

nskins:
I think it would make sense to use port 4321 here since we are instructed to pass in `PORT=4321` to start the TCP acceptor. This would probably be most natural since the user wants to ensure that environment variables are working as indicated in the guide.


josevalim:

PR #1617, opened 2022-07-10
Update dynamic-supervisor.markdown

thiagomajesk:
Ads a small note about the purpose of `Supervisor` and  `DynamicSupervisor` based on the previous improvements made https://github.com/elixir-lang/elixir/pull/11970 and https://github.com/elixir-lang/elixir/pull/11973.

### DIFF
--- a/getting-started/mix-otp/dynamic-supervisor.markdown
+++ b/getting-started/mix-otp/dynamic-supervisor.markdown
@@ -46,9 +46,7 @@ Since the bucket terminated, the registry also stopped, and our test fails when 
        test/kv/registry_test.exs:33: (test)
 ```
 
-We are going to solve this issue by defining a new supervisor that will spawn and supervise all buckets. Opposite to the previous Supervisor we defined, the children are not known upfront, but they are rather started dynamically. For those situations, we use a `DynamicSupervisor`. The `DynamicSupervisor` does not expect a list of children during initialization; instead each child is started manually via `DynamicSupervisor.start_child/2`.
-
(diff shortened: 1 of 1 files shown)


getting-started/mix-otp/dynamic-supervisor.markdown
```diff
@@ -48,6 +48,8 @@ Since the bucket terminated, the registry also stopped, and our test fails when
 
 We are going to solve this issue by defining a new supervisor that will spawn and supervise all buckets. Opposite to the previous Supervisor we defined, the children are not known upfront, but they are rather started dynamically. For those situations, we use a `DynamicSupervisor`. The `DynamicSupervisor` does not expect a list of children during initialization; instead each child is started manually via `DynamicSupervisor.start_child/2`.
 
+> It's important to notice that even though `Supervisor` is also capable of starting children dynamically, the `DynamicSupervisor` module allow us to rely on an efficient implementation designed for this specific purpose.
```
thiagomajesk, line 51: Should we exemplify what's special about DynSup? For instance, in the new docs, we mention very briefly that shutting down supervised processes can be done concurrently.

I'm not sure if should go for the extra information or just brush over the subject. Anyway, I'm not finding the right words to summarize it further so any suggestions are welcome. WDYT?

whatyouhide: I think this is looking good. I'd just slim down the phrasing a bit 🙃 

```suggestion
> Even though `Supervisor` is also capable of starting children dynamically, the `DynamicSupervisor` module allow us to rely on an efficient implementation designed for this specific purpose.   
```

whatyouhide (approved):

josevalim:

PR #1619, opened 2022-07-13
Add an additional note explaining what the '-S mix' flag is for

pepicrft:
### Context
I'm new to Elixir and while reading the documentation about Mix there was a code snippet that got me thinking about the purpose of the flag `-S mix` when invoking `iex`. I went straight to run `iex --help` but the output from the help, assuming that's the right way to get it as a user, lacked documentation that flag.

### What
I went ahead and added an additional phrase for people that might come across the same question. Please, let me know that's align how you'd expect this to be addressed or I should explore something else instead. For example improving the `iex --help` output. Thanks in advance for taking your time to review this tiny change.


josevalim:

PR #1620, opened 2022-07-16
Update wording to make clear @moduletag is for the whole module

nskins:
Updated "case" to "module" since `@moduletag :capture_log` refers to the whole module and not just a single test case.


josevalim:

PR #1621, opened 2022-07-20
Add "dbg" section in the "Debugging" guide

whatyouhide:


josevalim:
I think we can remove the IO.inspect section? If we want to keep it, I would have this new section after IO.inspect, as it is easier to use than Pry. :)

whatyouhide:
@josevalim ready for re-review 👀

josevalim:
I am thinking there is no benefit in talking about IEx.pry directly. We should just teach how to use it through “dbg” API and talk about breaks too. WDYT?

whatyouhide:
I don't have a strong preference. I like to talk about `pry` too because other folks might be more used to something like that. Maybe if the community moves to `dbg` more in the future, we can revisit this 🤔

josevalim:
>  I like to talk about pry too because other folks might be more used to something like that.

`pry` is very Ruby specific though and I think having too many options can actually be confusing because people won't know which one to reach out too. That's one of the points of `dbg`, so we should rally behind it.

whatyouhide:
@josevalim pushed more commits, ready for re-review! 🙃

josevalim (approved):
LGTM! :heart: however, we can only merge it once v1.14 is released. :)

josevalim:

PR #1625, opened 2022-08-12
corrected inconsistent iex prompts

dpreston:
Most of the corrections are just removing the line numbers from the `iex>` prompts to be consistent with the rest of the guide.
One is a trailing ) - `iex)>`
One is an empty trailing `iex>` prompt


josevalim:

PR #1626, opened 2022-08-12
File.read(file) receives a PID instead of a path

dpreston:
```elixir
{:ok, file} = File.open("hello", [:write])

case File.read(file) do
  {:ok, body}      -> # do something with the `body`
  {:error, reason} -> # handle the error caused by `reason`
end
```
If this example is used verbatim, you will get the error 
```
** (FunctionClauseError) no function clause matching in IO.chardata_to_string/1
    The following arguments were given to IO.chardata_to_string/1:
        # 1
        #PID<0.117.0>
    Attempted function clauses (showing 2 out of 2):
        def chardata_to_string(string) when is_binary(string)
        def chardata_to_string(list) when is_list(list)
    (elixir 1.13.4) lib/io.ex:627: IO.chardata_to_string/1
    (elixir 1.13.4) lib/file.ex:341: File.read/1
```
`file` should be a path, not the PID set earlier in the guide

### DIFF
--- a/getting-started/io-and-the-file-system.markdown
+++ b/getting-started/io-and-the-file-system.markdown
@@ -66,7 +66,7 @@ Notice that the version with `!` returns the contents of the file instead of a t
 The version without `!` is preferred when you want to handle different outcomes using pattern matching:
 
 ```elixir
-case File.read(file) do
+case File.read("hello") do
   {:ok, body}      -> # do something with the `body`
   {:error, reason} -> # handle the error caused by `reason`
 end
@@ -75,7 +75,7 @@ end
 However, if you expect the file to be there, the bang variation is more useful as it raises a meaningful error message. Avoid writing:
 
 ```elixir
-{:ok, body} = File.read(file)
+{:ok, body} = File.read("unknown")
 ```
 
 as, in case of an error, `File.read/1` will return `{:error, reason}` and the pattern matching will fail. You will still get the desired result (a raised error), but the message will be about the pattern which doesn't match (thus being cryptic in respect to what the error actually is about).
--- a/getting-started/io-and-the-file-system.markdown
+++ b/getting-started/io-and-the-file-system.markdown
@@ -34,13 +34,13 @@ hello world
 The [`File`](https://hexdocs.pm/elixir/File.html) module contains functions that allow us to open files as IO devices. By default, files are opened in binary mode, which requires developers to use the specific `IO.binread/2` and `IO.binwrite/2` functions from the `IO` module:
 
 ```elixir
-iex> {:ok, file} = File.open("hello", [:write])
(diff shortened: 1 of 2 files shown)


dpreston:
Urgh, accidently added the "try-catch-and-rescue.markdown" file as well to this patch which corrects an error where `what_happened` should not be defined but actually has been

getting-started/io-and-the-file-system.markdown
```diff
@@ -66,7 +66,7 @@ Notice that the version with `!` returns the contents of the file instead of a t
 The version without `!` is preferred when you want to handle different outcomes using pattern matching:
 
 ```elixir
-case File.read(file) do
+case File.read("hello") do
```
josevalim, line 69: Maybe?

```suggestion
case File.read("path/to/file") do
```

getting-started/io-and-the-file-system.markdown
```diff
@@ -75,7 +75,7 @@ end
 However, if you expect the file to be there, the bang variation is more useful as it raises a meaningful error message. Avoid writing:
 
 ```elixir
-{:ok, body} = File.read(file)
+{:ok, body} = File.read("unknown")
```
josevalim, line 78:
```suggestion
{:ok, body} = File.read("path/to/unknown-file")
```

?

dpreston: I think that's the clearer change, I just set them to use the previous example paths "hello" and "unknown" from the guide

josevalim: maybe we should update the previous values too then?

dpreston: Sounds good to me. There are a few variations throughout the guide:
Basic Types - "path/to/existing/file" and "path/to/unknown/file"
Enumerables - "path/to/file"
IO and the FS - "path/to/unknown"
All seem clearer that just "hello" and "unknown"

dpreston: IO and the file system also uses "hello" and "unknown"

josevalim: "path/to/existing/file" and "path/to/unknown/file" are the clearest to me but "path/to/file" and "path/to/unknown" works. Both are fine, your call :)

dpreston: "path/to/existing/file" and "path/to/unknown/file" sound good to me. What's the best message to update the PR?

Also, the change to what_happened in try-catch-and-rescue.markdown, any suggestions over what_else_happened?

josevalim: `another_what_happened` works for me!

dpreston (commented):
updated all path instances of `hello` and `unknown` for additional clarity

josevalim:

PR #1628, opened 2022-08-30
add Scoop install instructions for Windows

simonmcconnell:
chocolatey is rarely up to date and probably not even worth mentioning.  It probably does more harm than good having it here.


josevalim:

PR #1633, opened 2022-09-03
Link to correct precompiled binaries for supported OTP versions

kerryb:
The [existing installation instructions](http://elixir-lang.org/install.html#precompiled-package) point to an incorrect location for the precompiled binary download. This PR adds a separate link for each supported OTP version:

Screenshot 2022-09-03 at 17 30 37

I’ve added an `otp_versions` key to `_data/elixir-versions.yml`:

```yaml
v1_14:
  name: v1.14
  minimum_otp: 23.0
  otp_versions: [23, 24, 25]
  version: 1.14.0
  docs_zip: true
```

Not sure whether this is the best approach, but hopefully it’s a good start.

### DIFF
--- a/_data/elixir-versions.yml
+++ b/_data/elixir-versions.yml
@@ -3,6 +3,7 @@ stable: v1_14
 v1_14:
   name: v1.14
   minimum_otp: 23.0
+  otp_versions: [23, 24, 25]
   version: 1.14.0
   docs_zip: true
 
--- a/install.markdown
+++ b/install.markdown
@@ -125,7 +125,12 @@ The above will automatically point to the latest Erlang and Elixir available. Fo
 
 ## Precompiled package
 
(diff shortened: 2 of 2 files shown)


Eiji7:
Looks good. I'm not sure if we should have so much space between list items … For me it looks like 3 lists each having one item, but someone may say it looks good as is …

kerryb:
> @kerryb Looks good. I'm not sure if we should have so much space between list items … For me it looks like 3 lists each having one item, but someone may say it looks good as is …
> …
> I may only guess, but for me it looks like the template generates extra empty lines around each `* list element` making it a separate list with only one item.

Yeah, I couldn’t figure out originally how to get it to generate proper markdown, but it turns out you just need to move the `{% endfor %}` to the end of the previous line.

Branch updated.

Eiji7 (approved):

josevalim:

PR #1635, opened 2022-09-06
Add example with multiple nesting to `alias` guide

madkinsz:
Updates the getting started guide for nested aliases to clarify `alias` behavior when applied to a module with multiple nesting e.g. `alias Foo.Bar.Baz`.

I think find this behavior very logical, but I was not certain what behavior Elixir provided without checking in `iex`.


josevalim:

PR #1636, opened 2022-09-06
Add `use` to sentence in multi-alias example

madkinsz:
This sentence was missing "use" despite its inclusion in the heading. The change includes the addition of code formatting to these words for clarification so it doesn't read as "... or use multiple modules".

If you'd prefer a roll-up of any changes I find in the getting started guide I can do so. It seems easier to accept/deny them separately though.

### DIFF
--- a/getting-started/alias-require-and-import.markdown
+++ b/getting-started/alias-require-and-import.markdown
@@ -219,7 +219,7 @@ As we will see in later chapters, aliases also play a crucial role in macros, to
 
 ## Multi alias/import/require/use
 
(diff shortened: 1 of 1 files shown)


madkinsz:
Interestingly this chapter is also titled "alias, require, and import". Perhaps this should be updated to include "use"?

It may make sense to retitle this entirely as "Using modules"?

josevalim:

josevalim:
The reason we don't mention `use` in the title is because it is really just a `require` + macro. It is not a special form per-se, so we did not give it a "special" call. :)

madkinsz:
Ah great that makes sense, thanks for explaining! It may be worth changing the title to be more general since it's in the Getting started guide and new users are probably ignorant of language details.

Similarly, https://elixir-lang.org/getting-started/case-cond-and-if.html could be "Control flow" and https://elixir-lang.org/getting-started/try-catch-and-rescue.html could be "Error handling" for consistency with the rest of the headers.

I don't have strong feelings here but hope my perspective as a newcomer to the documentation can be helpful :) If you want me to open a pull request let me know.

PR #1638, opened 2022-09-21
Format and clean up crash-course

doughsay:
A friend of mine recently found this page, and when he asked me questions about it I saw that it was pretty outdated. I've gone and updated the formatting of at least all the Elixir code to match the modern formatter and idioms, but I'm not familiar enough with Erlang to know if those examples need updating or not.

Hope this helps, please let me know if I should change anything.

### DIFF
--- a/crash-course.markdown
+++ b/crash-course.markdown
@@ -28,7 +28,7 @@ hello() ->
 Add your functions to it, save it to disk, run `erl` from the same directory and execute the `compile` command:
 
 ```erl
-Eshell V5.9  (abort with ^G)
+Eshell V13.0.4  (abort with ^G)
 1> c(module_name).
 ok
 1> module_name:hello().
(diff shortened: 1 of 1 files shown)


crash-course.markdown
```diff
@@ -431,13 +438,14 @@ end
...
 
-HelloModule.dummy
+HelloModule.dummy()
 #=> :ok
 
-HelloModule.Utils.util
-#=> "Utilize"
+HelloModule.Utils.util()
+# "Utilize"
+#=> :ok
```
doughsay, line 446: Wasn't quite sure how to show here that there's stdout printing in addition to a return value.

crash-course.markdown
```diff
@@ -614,11 +622,12 @@ F({a, b}).
...
-        IO.puts "All your #{inspect tuple} are belong to us"
-      [] ->
-        "Empty"
-    end
+  {:a, :b} = tuple ->
+    "All your #{inspect(tuple)} are belong to us"
+
+  [] ->
+    "Empty"
+end
```
doughsay, line 630: Removing the IO.puts feels right to me here, but it causes it to differ from the Erlang example. I don't know how to change the Erlang example to make it match.

josevalim:

PR #1640, opened 2022-10-01
Redirect http traffic to https

jirkavrba:
When visiting elixir-lang.org I noticed that the there is no automatic redirect to https so I added a small script to perform such redirect when the http:// version of the page is loaded


josevalim:

PR #1652, opened 2022-11-05
Update the "Supervisor and Application" chapter

artnest:
Update output on child process crash to match the current Elixir version.


josevalim:

PR #1668, opened 2023-02-03
Update keywords-and-maps.markdown

daniel-cavalcante:
The phrase "In particular, keyword lists are 2-item tuples where the first element ..." states that a keyword list *is* a 2-item tuple. As someone who is just reading about such concepts for the first time, I have found it to be a source of confusion, because in reality a keyword list is a list where each element is a 2-item tuple (this is accurate).


josevalim:

PR #1669, opened 2023-02-03

kacorvus:
It's easy to forget to run the doc tests. It would be nice to emphasis this point more strongly.


josevalim:
Thank you @kacorvus but I'd prefer to not go ahead with this for now. Pretty much many of the steps in all of the guides are necessary and lead to different behaviour if we don't do them, so I am not sure we should call this one out specifically. And hopefully the example right below it helps too. :) Thanks anyway!

josevalim:
Got it, I will push a comment earlier. Thanks for given more context!

PR #1673, opened 2023-03-07
Fix version check command

Nicd:
The current command proposed for checking the Erlang version is `erl -s halt`, but that makes `init` call the module `halt`, which does not exist, and it results in a crash. This may mislead users into thinking their Erlang version is not working properly.

In this change the module name is added so that it calls `erlang:halt` correctly. This results in no crash.

Tested on Erlang 25 and Elixir 1.14. Checked in the Erlang docs that `-s` has worked this way (`-s Mod Fun`) since Erlang 18 at least.


josevalim:

PR #1677, opened 2023-03-12
Update dependencies-and-umbrella-projects.markdown

simonelnahas:
I'm not sure "build the intuition" is the right phrase to use in this context.

### DIFF
--- a/getting-started/mix-otp/dependencies-and-umbrella-projects.markdown
+++ b/getting-started/mix-otp/dependencies-and-umbrella-projects.markdown
@@ -198,7 +198,7 @@ deps_path: "../../deps",
 lockfile: "../../mix.lock",
 ```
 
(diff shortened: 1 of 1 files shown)


getting-started/mix-otp/dependencies-and-umbrella-projects.markdown
```diff
@@ -198,7 +198,7 @@ deps_path: "../../deps",
 lockfile: "../../mix.lock",
 ```
 
-Those options mean all dependencies will be checked out to `kv_umbrella/deps`, and they will share the same build, config and lock files. We haven't talked about configuration yet, but from here we can build the intuition that all configuration and dependencies are shared across all projects in an umbrella, and it is not per application.
+Those options mean all dependencies will be checked out to `kv_umbrella/deps`, and they will share the same build, config and lock files. We haven't talked about configuration yet, but from here we can get the idea that all configuration and dependencies are shared across all projects in an umbrella, and it is not per application.
```
josevalim, line 201:
```suggestion
Those options mean all dependencies will be checked out to `kv_umbrella/deps`, and they will share the same build, config, and lock files. We haven't talked about configuration yet, but keep in mind all configuration and dependencies are shared across all projects in an umbrella, and it is not per application.
```

josevalim:

PR #1680, opened 2023-03-16
Suggestion for additional example for Macro.escape

simonelnahas:
This made it easier to understand for me since we can directly compare the functionality. What do you think?

I don't have time for reformulating the text right now, but it could be considered in the future.


getting-started/meta/quote-and-unquote.markdown
```diff
@@ -146,6 +146,8 @@ In some cases, you may need to inject such *values* into *quoted expressions*. T
 
 ```elixir
 iex> map = %{hello: :world}
+iex> quote do: map
+{:m, [], Elixir}
```
josevalim, line 150:
```suggestion
{:map, [], Elixir}
```

whatyouhide:
Thanks @simonelnahas! 💟

Issue #1684, opened 2023-03-27
Small reordering of information in "alias, require, and import"


siiky:
Hi,

In the "alias, require, and import" chapter of the Getting Started guide, the order in which macros are referenced is not so natural, as in, why are macros mentioned at all in the sentence "means that ..." ? I think it would be better if the last sentence stating that importing also requires was close to the top, before the other one.

> Note that `import` is **lexically scoped** too. This means that we can import specific macros or functions inside function definitions:
> (...)
> `import`ing a module automatically `require`s it.

https://github.com/elixir-lang/elixir-lang.github.com/blob/86b983f0c38fa1810a538a7d2ccecc28bb471bdb/getting-started/alias-require-and-import.markdown?plain=1#L106-L117

PR #1685, opened 2023-03-27
Reorder `require` mention in `import` section

siiky:
Related: https://github.com/elixir-lang/elixir-lang.github.com/issues/1684

I re-read the first sentence of the section and noticed that macros are indeed mentioned (I must've not noticed it yesterday when I read it). Still, I think it could/should be made more explicit right at the beginning.

WDYT?

### DIFF
--- a/getting-started/alias-require-and-import.markdown
+++ b/getting-started/alias-require-and-import.markdown
@@ -90,7 +90,7 @@ Note that like the `alias` directive, `require` is also lexically scoped. We wil
 
 ## import
 
-We use `import` whenever we want to access functions or macros from other modules without using the fully-qualified name. Note we can only import public functions, as private functions are never accessible externally.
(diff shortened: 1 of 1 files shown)


getting-started/alias-require-and-import.markdown
```diff
@@ -90,7 +90,7 @@ Note that like the `alias` directive, `require` is also lexically scoped. We wil
 
 ## import
 
-We use `import` whenever we want to access functions or macros from other modules without using the fully-qualified name. Note we can only import public functions, as private functions are never accessible externally.
+We use `import` whenever we want to access functions or macros from other modules without using the fully-qualified name. Note we can only import public functions, as private functions are never accessible externally, and that `import`ing a module automatically `require`s it.
```
josevalim, line 93:
```suggestion
We use `import` whenever we want to access functions or macros from other modules without using the fully-qualified name. Note we can only import public functions, as private functions are never accessible externally. `import`ing a module automatically `require`s it.
```

josevalim:

PR #1690, opened 2023-04-10
Revise last paragraph in "Optional Syntax"

rhcarvalho:
An attempt to make it better.

- I think "worried" conveys fear, and we can replace it with "concerned" for a more neutral tone
- We can't really *ensure* universal consistency, but provide standard tools with standard config suggested for most uses


josevalim:

PR #1691, opened 2023-04-11
Update typespecs-and-behaviours.markdown w/consistent variable name in example

stevensonmt:
Make variable for file extension consistent across with clauses in behaviour use example.


whatyouhide:
Great catch! 💯 Thanks @stevensonmt 💟

PR #1694, opened 2023-04-18
Update config-and-releases.markdown

GuiHeurich:
When `MIX_ENV=prod mix release foo` is run, the output doesn't include an exclamantion mark at the end.


josevalim:

Virginboy202:
> When `MIX_ENV=prod mix release foo` is run, the output doesn't include an exclamantion mark at the end.

PR #1699, opened 2023-06-12

David-Tennant:
The "last" topic feels misleading. I don't like "next" topic either but im not sure if these sentence was meant to be "last topic with regards to modules" or "last topic in the guide"


josevalim:

PR #1700, opened 2023-06-14
Update structs.markdown

David-Tennant:
Removed ambiguity

Unsure what `(and *all* of them)` means. Possibly in an older version of Elixir you had to pass in all the properties to a struct?


josevalim:

sabiwara:
I think the intent was probably to say it also ensures all keys are present (even when not passed):

```elixir
iex> user = %User{name: "Joe"}
iex> user.age
27
```

I'm not sure what would be a better way of saying it though. Maybe a separate sentence with a separate example?

David-Tennant:
@sabiwara that does sound like a reasonable assumption. Above it mentions there are default values that are provided. 

```
The keyword list used with `defstruct` defines what fields the struct will have along with their default values.
```

> Maybe a separate sentence with a separate example?

That would be great

PR #1708, opened 2023-08-04
Correct command to run tests in “Distributed tasks and tags” chapter

knowler:
Just noticed this while going through the guide.


josevalim:
Thank you. However both commands should be fine, it depends if you want an IEx shell attached or not. :)

PR #1710, opened 2023-08-08
add in change to allow a charlist to be displayed as a list of integers

magashpillay:
add in change to allow a charlist to be displayed as a list of integers


josevalim:
We are moving these guides to elixir-lang/elixir, can you please update it there?

magashpillay:
I am going to give it a try- I am very new to Elixir and finding my way around

PR #1715, opened 2023-08-19
Update undefined variable error messages

sabiwara:
Backporting https://github.com/elixir-lang/elixir/pull/12876


whatyouhide:
Nice, thanks @sabiwara! ❤️

PR #1719, opened 2023-08-28
Add hiring section?

pdgonzalez872:
Hi!

I posted on [ElixirForum](https://elixirforum.com/t/hiring-section-maybe/57883) about this. It seems folks think it's an ok idea? 11 likes at the moment. Maybe they are just trying to be nice 🙂 

After chatting with some friends about this, I'm still not convinced this concept fits here on the website even though the intent is a good one and the problem does seem to exist.

I thought I should move this here since I don't expect folks to be watching everything that happens on EF. Feel free to do whatever you'd like with this: close the PR (no hard feelings at all), change wording/tone, change sentences, anything you'd see fit.

As usual, thank you so much for all of your hard work! ❤️

### DIFF
--- a/_includes/important-links.html
+++ b/_includes/important-links.html
@@ -27,6 +27,7 @@
     <li><a class="spec" href="https://www.meetup.com/topics/elixir/">Meetups around the world</a></li>
     <li><a class="spec" href="https://github.com/elixir-lang/elixir/wiki">Wiki with events and resources maintained by
         the community</a></li>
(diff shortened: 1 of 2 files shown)


pdgonzalez872:
unless you want to keep the link in `important_links` and point it to the wiki? I'm not sure where you meant the link would live. Anywhere you see fit!

josevalim:
Let's keep the sidebar link and point to the wiki, yes.

PR #1725, opened 2023-09-20
New blog post on strong arrows

josevalim:


_posts/2023-09-20-strong-arrows-gradual-typing.markdown
```diff
@@ -0,0 +1,281 @@
+---
+layout: post
+title: "Strong arrows: a new approach to gradual typing"
+author: José Valim
+category: Announcements
+excerpt: A short status update on the effort to bring a type system into Elixir.
```
josevalim, line 6:
```suggestion
excerpt: An introduction to strong arrows and how it leverages the Erlang VM to provide sound gradual typing.
```

_posts/2023-09-20-strong-arrows-gradual-typing.markdown
```diff
@@ -0,0 +1,281 @@
...
+```elixir
+$ a -> a when a: number()
+def identity(arg), do: arg
+```
+
+But at the end of the day it will simply expand to intersections. The important bit is that, at the semantic level, there is no need for additional constructs and representations.
+
+> Note: for the type-curious readers, set-theoretic types implement a limited form of bounded quantification a-la kernel fun. In a nutshell, it means we can only compare functions if they have the same bounds. For example, our type system cannot answer if `a -> a when a: integer() or boolean()` is a subtype of `a -> a when a: integer()`.
+
+Not only that, we get lower bounds for free. If intersections allow us to place an upper bound, a union is a lower bound, because it says a given type must always be considered: `a or atom()` means atoms and any other type (which may or may not be a subset of atoms).
```
sabiwara, line 104: This transition is a bit brutal, especially because this whole section is about intersections and constraints, except this block (which is followed by another sentence about intersections). Also, the use case isn't very clear.

Maybe developing a more concrete example could help?
For example, I could imagine `a or nil` as an equivalent of an option type? (don't know if my understanding is correct)

_posts/2023-09-20-strong-arrows-gradual-typing.markdown
```diff
@@ -0,0 +1,281 @@
...
+## Introducing strong arrows
+
+I have always said that Elixir, thanks to Erlang, is an assertive language. For example, if our identity function is restricted to only numbers, in practice we would most likely write it as:
+
+```elixir
+$ a -> a when a: number()
+def identity(arg) when is_number(arg), do: arg
+```
+
+In the example above, `identity` will fail if given any value that is not an integer. We often rely on pattern matching and guards and, in turn, they helps us assert on the types we are working with. Not only that, Erlang's JIT compiler already relies on this information to perform optimizations whenever possible.
```
sabiwara, line 168: typo + maybe provide a link about JIT optimizations:

```suggestion
In the example above, `identity` will fail if given any value that is not a number. We often rely on pattern matching and guards and, in turn, they helps us assert on the types we are working with. Not only that, Erlang's JIT compiler already relies on this information to [perform optimizations](https://www.erlang.org/blog/type-based-optimizations-in-the-jit/) whenever possible.
```

_posts/2023-09-20-strong-arrows-gradual-typing.markdown
```diff
@@ -0,0 +1,281 @@
...
+
+We also say Elixir is strongly typed because its functions and operators avoid implicit type conversions. The following functions also fail when the input does not match its type:
```
sabiwara, line 170:
```suggestion
We also say Elixir is strongly typed because its functions and operators avoid implicit type conversions. The following functions also fail when their input does not match their type:
```

_posts/2023-09-20-strong-arrows-gradual-typing.markdown
```diff
@@ -0,0 +1,281 @@
...
+2
+iex(2)> increment(13.0)
+14.0
+iex(3)> increment("foobar")
+** (ArithmeticError) bad argument in arithmetic expression: "foobar" + 1
+```
+
+In other words, Elixir's runtime consistently checks the values and their types at runtime. If `increment` fails when given something else than a number, then it will fail when the `dynamic()` type does not match its input at runtime. This guarantees `increment` returns its declared type and therefore we do not need to introduce runtime type checks when calling said function from untyped code.
+
+When we look at the `identity`, `debug`, and `increment` functions above, we - as developers - can state that these functions raise when given a value that does not match its input. However, how can we generalize this property so it is computed by the type system itself? To do so, we introduce a concept called **strong arrows**, which rely on set-theoretical types to derive this property.
```
sabiwara, line 193:
```suggestion
When we look at the `identity`, `debug`, and `increment` functions above, we - as developers - can state that these functions raise when given a value that does not match their input. However, how can we generalize this property so it is computed by the type system itself? To do so, we introduce a concept called **strong arrows**, which rely on set-theoretical types to derive this property.
```

_posts/2023-09-20-strong-arrows-gradual-typing.markdown
```diff
@@ -0,0 +1,281 @@
...
+
+There is one last scenario we must take into consideration when interfacing dynamic and static code. Imagine the following code:
+
+```elixir
+def increment_and_remainder(numerator, denominator) do
+  rem(numerator, increment(denominator))
+end
+```
+
+The function receives two dynamically typed arguments and computes the remainder of the numerator by the denominator incremented by one. The `rem/2` operation expects both sides to be integers and returns an integer. Therefore the type of `increment_and_remainder/2` would be `dynamic(), dynamic() -> integer()`. Let's also consider that this function is part of an existing codebase where all calls to `increment_and_remainder` pass two valid integers.
```
sabiwara, line 251: Sorry for this probably silly question, but why isn't the type of `numerator` `dynamic() and integer()` or even `integer()`, since anything else is bound to fail?

_posts/2023-09-20-strong-arrows-gradual-typing.markdown
```diff
@@ -0,0 +1,281 @@
...
+
+> Errata: my presentation used the type `integer()` instead of `number()` for this example. However, that was a mistake in the slide. Giving the type `integer(), integer() -> integer()` to `subtract` and `integer() -> integer()` to `negate` does not make `subtract` a strong arrow. Can you tell why?
+
+Luckily, strong arrows can also be leveraged by other gradually typed languages. However, the more polymorphic a language and its functions are, the more unlikely it is to conclude that a given function is strong. For example, in other gradually typed languages such as Python or Ruby, the `+` operator is extensible and the user can define custom types where the operation is valid. In TypeScript, `"foobar" + 1` is also a valid operation, which expands the function domain. In both cases, an `increment` function restricted to numbers would not be receive a strong arrow type. Therefore, to remain sound, they must either restrict the operands with further runtime checks or return `dynamic()` and reduce the number of compile-time checks.
```
sabiwara, line 228:
```suggestion
Luckily, strong arrows can also be leveraged by other gradually typed languages. However, the more polymorphic a language and its functions are, the more unlikely it is to conclude that a given function is strong. For example, in other gradually typed languages such as Python or Ruby, the `+` operator is extensible and the user can define custom types where the operation is valid. In TypeScript, `"foobar" + 1` is also a valid operation, which expands the function domain. In both cases, an `increment` function restricted to numbers would not be considered a strong arrow type. Therefore, to remain sound, they must either restrict the operands with further runtime checks or return `dynamic()` and reduce the number of compile-time checks.
```

_posts/2023-09-20-strong-arrows-gradual-typing.markdown
josevalim: There aren't many examples of lower bounds. Perhaps I should move it to a "note", like I did above? It is not really required for the understanding of the text.

_posts/2023-09-20-strong-arrows-gradual-typing.markdown
josevalim: You are thinking of typing inference: the ability to look at how something is used in the code to deduce its type. We are not doing inference though, we are doing type checking.

_posts/2023-09-20-strong-arrows-gradual-typing.markdown
```diff
@@ -0,0 +1,285 @@
...
+```elixir
+$ a -> a when a: number()
+def identity(arg), do: arg
+```
+
+But at the end of the day it will simply expand to intersections. The important bit is that, at the semantic level, there is no need for additional constructs and representations.
+
+> Note: for the type-curious readers, set-theoretic types implement a limited form of bounded quantification a-la kernel fun. In a nutshell, it means we can only compare functions if they have the same bounds. For example, our type system cannot answer if `a -> a when a: integer() or boolean()` is a subtype of `a -> a when a: integer()`.
+
+Not only that, we get lower bounds for free. If intersections allow us to place an upper bound, a union is a lower bound, because it says a given type must always be considered: `a or atom()` means atoms and any other type (which may or may not be a subset of atoms).
```
josevalim, line 104:
```suggestion
We also get lower bounds for free. If intersections allow us to place an upper bound on a type variable, a union is equivalent to a lower bound, because it says the type variable will always be augmented by the union-ed type. For example, `a or atom()` says the result will always include atoms plus whatever else specified by `a` (which may be an atom, `atom()` itself, or a completely disjoint type such as `integer()`).
```

_posts/2023-09-20-strong-arrows-gradual-typing.markdown
```diff
@@ -0,0 +1,285 @@
...
+
+```elixir
+def increment_and_remainder(numerator, denominator) do
+  rem(numerator, increment(denominator))
+end
+```
+
+The function receives two dynamically typed arguments and computes the remainder of the numerator by the denominator incremented by one. The `rem/2` operation expects both sides to be integers and returns an integer. Therefore the type of `increment_and_remainder/2` would be `dynamic(), dynamic() -> integer()`. Let's also consider that this function is part of an existing codebase where all calls to `increment_and_remainder` pass two valid integers.
+
+Here lies the issue: if `increment(dynamic())` returns `number()`, then `number()` includes both `integer() or float()`, and therefore the program above won't type check because `rem/2` does not accept floats. When faced with this problem, there are two possible reactions:
```
josevalim, line 257:
````suggestion

$ (integer() -> integer()) and (float() -> float())
def increment(number), do: number + 1
```

The `increment_and_remainder/2` function is untyped, therefore both of its arguments receive type `dynamic()`. The function then computes the remainder of the numerator by the denominator incremented by one. For this example, let's assume all uses of `increment_and_remainder/2` in our program passes two integers as arguments.

Given `increment/1` has a strong arrow type, according to our definition, `increment(dynamic())` will return `integer() or float()` (also known as `number()`). Here lies the issue: if `increment(dynamic())` returns `integer() or float()`, the program above won't type check because `rem/2` does not accept floats.

When faced with this problem, there are two possible reactions:
````

File: _posts/2023-09-20-strong-arrows-gradual-typing.markdown [hunk 10]
```diff
@@ -0,0 +1,290 @@
...
+
+Given `increment/1` has a strong arrow type, according to our definition, `increment(dynamic())` will return `integer() or float()` (also known as `number()`). Here lies the issue: if `increment(dynamic())` returns `integer() or float()`, the program above won't type check because `rem/2` does not accept floats.
+
+When faced with this problem, there are two possible reactions:
+
+1. It is correct for the function to not type check given increment may return a float
+
+2. It is incorrect for the function to not type check because the error it describes never occurs in the codebase
+
+Another interesting property of gradual set-theoretic types is that we can also define upper bounds on the `dynamic()` type. If a function returns `number()`, it means the caller needs to handle both `integer()` and `float()`. However, if a function returns `dynamic() and number()`, it means the type is defined at runtime, but it must still verify it is one of `integer()` or `float()` at compile time. Therefore, the return of `dynamic() and number()` would not emit false positives, but if you attempt to use the string concatenation operator (`<>`) on it, you'd still get a typing violation!
```
josevalim, line 268:
```suggestion
Another interesting property of gradual set-theoretic types is that we can also define upper bounds on the `dynamic()` type. If a function returns `number()`, it means the caller needs to handle both `integer()` and `float()`. However, if a function returns `dynamic() and number()`, it means the type is defined at runtime, but it must still verify it is one of `integer()` or `float()` at compile time.

Therefore, `rem/2` will type check if its second argument has the type `dynamic() and number()`, as there is one type at runtime (`integer()`) that satisfies type checking. On the other hand, if you attempt to use the string concatenation operator (`<>`) on `dynamic() and number()`, then there is no acceptable runtime type and you'd still get a typing violation!
```

_posts/2023-09-20-strong-arrows-gradual-typing.markdown
```diff
@@ -0,0 +1,292 @@
...
+Or course, we can provide syntax sugar for those constraints:
+
+```elixir
+$ a -> a when a: number()
+def identity(arg), do: arg
+```
+
+But at the end of the day it will simply expand to intersections. The important bit is that, at the semantic level, there is no need for additional constructs and representations.
+
+> Note: for the type-curious readers, set-theoretic types implement a limited form of bounded quantification a-la kernel fun. In a nutshell, it means we can only compare functions if they have the same bounds. For example, our type system cannot answer if `a -> a when a: integer() or boolean()` is a subtype of `a -> a when a: integer()`.
```
josevalim, line 102:
```suggestion
> Note: for the type-curious readers, set-theoretic types implement [a limited form of bounded quantification a-la Kernel Fun](http://lucacardelli.name/Papers/OnUnderstanding.pdf). In a nutshell, it means we can only compare functions if they have the same bounds. For example, our type system cannot answer if `a -> a when a: integer() or boolean()` is a subtype of `a -> a when a: integer()`.
```

PR #1728, opened 2023-10-01
Clarify Any Protocol implementation derivation

Stevo-S:
A small edit on the "Deriving" section of the "Protocols" chapter in "Getting Started".

Changing "we should be fine with the implementation of Any" to
"should we be fine with the implementation of any" hopefully
better communicates that we can use the Any implementation of
the protocol if we wish to but not that we have to.

### DIFF
--- a/getting-started/protocols.markdown
+++ b/getting-started/protocols.markdown
@@ -166,7 +166,7 @@ end
 
 The implementation above is arguably not a reasonable one. For example, it makes no sense to say that the size of a `PID` or an `Integer` is `0`.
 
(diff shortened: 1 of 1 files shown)


josevalim:
Hi @Stevo-S! The docs have been moved to Elixir repo itself, could you please submit the patch there instead? We will swap once a new version is out. You can look for a file of the same name.

Stevo-S:
I've created a [PR ](https://github.com/elixir-lang/elixir/pull/12978) on the Elixir repo as suggested.

Closing this one.

PR #1735, opened 2023-10-21
Explicitly use 127.0.0.1 as IP address

adrianbrink:
On macOS `gen_tcp.listen()` wouldn't work without passing the explicit address for localhost. KVServer would start correctly, but the Telnet client wouldn't connect.


0dayfall (approved):

josevalim:
Interesting. I have always ran this on macOS and it has always worked. I just tested it again and it is fine. Which version are you using?

josevalim:
Closing this for now. Note if the issue persists, the guides have been moved to elixir repository itself. :) Thanks!

PR #1737, opened 2023-10-25
Fix incorrect byte_size example

tonydangblog:
The example in [Getting Started - Binaries, strings, and charlists](https://elixir-lang.org/getting-started/binaries-strings-and-char-lists.html#utf-8-and-encodings) shows `byte_size("héllo")` returning `6`, but it should return `7` since the character `é` is three bytes.

Current snippet:
```zsh
iex> string = "héllo"
"héllo"
iex> String.length(string)
5
iex> byte_size(string)
6
```

Proposed change:
```zsh
iex> string = "héllo"
"héllo"
iex> String.length(string)
5
iex> byte_size(string)
7 # <- should return 7 here
```

### DIFF
--- a/getting-started/binaries-strings-and-char-lists.markdown
+++ b/getting-started/binaries-strings-and-char-lists.markdown
@@ -53,10 +53,10 @@ iex> string = "héllo"
 iex> String.length(string)
 5
 iex> byte_size(string)
-6
+7
 ```
 
-Although the string above has 5 characters, it uses 6 bytes, as two bytes are used to represent the character `é`.
+Although the string above has 5 characters, it uses 7 bytes, as three bytes are used to represent the character `é`.
 
 > Note: if you are running on Windows, there is a chance your terminal does not use UTF-8 by default. You can change the encoding of your current session by running `chcp 65001` before entering `iex` (`iex.bat`).
 


josevalim:
It depends. "é" can be written in two bytes as a single character "é" or three bytes as the character "e" followed by the accent. Which operating system and terminal are you using?

tonydangblog:
Oh, I didn't know that! I'm using the default Terminal on Mac with zsh.

Screenshot 2023-10-25 at 15 53 42

josevalim:
Interesting. So maybe it is your keyboard settings? What is your default keyboard layout? Or more precisely, how are you typing the `é`?

tonydangblog:
I'm not typing the `é`. I'm copying and pasting it directly from the example in the Getting Started guide.

tonydangblog:
I just tried typing the `é` with my keyboard instead and it returned as two bytes! So looks like it's definitely something to do with the copy & pasting!

tonydangblog:
Hmm, to avoid confusion like this, what do you think about changing the example to use `hellö` instead of `héllo`? I did not run into the same problem when copy&pasting `hellö`.

josevalim:
Interesting. In my browser, "héllo" has six bytes on Firefox but 7 on Safari, so Safari is the one normalizing it! I wonder if this is recent, because otherwise we would have heard about it sooner.

However, we are currently moving the docs to Elixir: https://hexdocs.pm/elixir/main/binaries-strings-and-charlists.html - and in this page, it has 6 on *both* browsers. So it is definitely a combination of both. By the end of next week latest, I will update all links to point to the new docs, so this will be resolved then.

Btw, unfortunately "hellö" could have the same issue, so if the issue comes back, we would need to pick something else altogether, such as emoji.

tonydangblog:
Nice, I just checked the new docs and can confirm copy&pasting results in 6 bytes using Chrome, Firefox, and Safari. I'll go ahead and close this PR.

Thanks for taking the time to look at this small issue. This is my first time looking at Elixir so I'm glad the community is very responsive!

PR #1743, opened 2023-12-23
Fix meta programming anti-patterns link

PragTob:
The link in the release blog post was broken, fixed with the current right one - whose name is macro not meta :)

Thanks for the 1.16 release and happy holidays!


sabiwara (approved):
Thank you @PragTob ! 💜

PR #1747, opened 2024-01-25
Set og:image meta for seo by default on all pages

sorentwo:
Links to the site don't generate a "social" preview without an `og:image` or `twitter:image` tag. This adds a larger variant of the horizontal logo and sets it as the default social image. It may be nice to set per-page social images for blog posts and cases, but I didn't tackle that in this PR.

One unrelated addition here is replacing the pixelated RSS feed `png` with a scalable `svg` because it neighbored the new `elixir-og-card.jpg`.

How to set images is described in ["Advanced Usage" in the jekyll-seo reference](https://github.com/jekyll/jekyll-seo-tag/blob/master/docs/advanced-usage.md).


josevalim:

josevalim:
Unfortunately I had to revert this PR because it broke all blog links. For example, it was linking to https://elixir-lang.org/releases/elixir-v1-16-0-released.html instead of  /blog/2023/.... 

Can you please take a look and resubmit @sorentwo?

PR #1748, opened 2024-02-04
Specify og:image for posts and individual pages

sorentwo:
This addresses the reversion of https://github.com/elixir-lang/elixir-lang.github.com/pull/1747

Now the permalink format remains set for all pages and we add the image to each separate template. It's more verbose, but easier to reason about and none of the links break 🙂


josevalim:

PR #1752, opened 2024-03-01
Expand install instructions for Debian

tzugen:
On Debian bookworm the package erlang-dev needs to be installed as well, otherwise you will have problems creating the first elixir application (see https://github.com/elixir-lang/elixir/issues/12681#issuecomment-1699374362 )

### DIFF
--- a/install.markdown
+++ b/install.markdown
@@ -53,7 +53,7 @@ Keep in mind that each Elixir version supports specific Erlang/OTP versions. [Se
     * Run: `pacman -S elixir`
 
   - **Debian**
-    * Run: `sudo apt install elixir`. If you plan to develop Elixir applications you need `erlang-dev` as well. 
+    * Run: `sudo apt install erlang-dev elixir`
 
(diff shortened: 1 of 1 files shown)


install.markdown
```diff
@@ -53,7 +53,7 @@ Keep in mind that each Elixir version supports specific Erlang/OTP versions. [Se
     * Run: `pacman -S elixir`
 
   - **Debian**
-    * Run: `sudo apt install elixir`
+    * Run: `sudo apt install elixir`. If you plan to develop Elixir applications you need `erlang-dev` as well.
```
josevalim, line 56: What do you think about suggesting both? Does it work?

```suggestion
    * Run: `sudo apt install erlang-dev elixir`
```

tzugen: It makes more sense to be honest :)

tzugen: And it works

josevalim:

PR #1754, opened 2024-03-03
Fix Erlang version of sum/2 function in the crash course article

7006:
I've copied Erlang version of function sum/2 from https://elixir-lang.org/crash-course.html#identifying-functions, compiled and run it in the Erlang shell

The "a" in Erlang is actually a list(yep it's weird) so we fall into the 2nd clause of sum/2 function

So we should use binary syntax in order to fall into the 3rd clause of function

Screenshot 2024-03-04 at 00 16 28

Screenshot 2024-03-04 at 00 17 29


sabiwara (approved):
Good catch, thank you @7006 💜

whatyouhide:
Thanks @7006! 💟

PR #1757, opened 2024-05-04
Fix Nerves case study link

osbre:


whatyouhide:
Good catch, thanks @osbre 💟

PR #1759, opened 2024-05-14
Document installing recent releases from Fedora Rawhide

rhcarvalho:
Those using Fedora can get access to more recent releases, often bleeding edge, using Fedora's package manager.

Also note that documentation comes in a separate package, and can be useful for offline access.


josevalim:

PR #1768, opened 2024-08-31
Update `String.split/2` into `String.split/3`

zorn:
There is no `String.split/2`, best I can tell.

https://hexdocs.pm/elixir/String.html#split/3

But `String.split/3` does behave as the blog post explains.

```
iex(1)> String.split("")
[]
iex(2)> String.split("", ",")
[""]
```

### DIFF
--- a/_posts/2024-08-28-typing-lists-and-tuples.markdown
+++ b/_posts/2024-08-28-typing-lists-and-tuples.markdown
@@ -97,7 +97,7 @@ While different developers will prefer certain idioms over others, I am not conv
 
 ### What about Elixir?
 
-Thanks to set-theoretic types, we will most likely distinguish between empty lists and non-empty lists in Elixir's type system, since pattern matching on them is a common language idiom. Furthermore, several functions in Elixir, such as `String.split/2` are guaranteed to return non-empty lists, which can then be nicely encoded into a function's return type.
+Thanks to set-theoretic types, we will most likely distinguish between empty lists and non-empty lists in Elixir's type system, since pattern matching on them is a common language idiom. Furthermore, several functions in Elixir, such as `String.split/3` are guaranteed to return non-empty lists, which can then be nicely encoded into a function's return type.
 
 Elixir also has the functions `hd` (for head) and `tl` (for tail) inherited from Erlang, which are [valid guards](https://hexdocs.pm/elixir/patterns-and-guards.html). They only accept non-empty lists as arguments, which will now be enforced by the type system too.
 


josevalim:
That's a bug in Elixir. I will take a look at it. :) Thanks for noticing.

sabiwara:
> There is no String.split/2, best I can tell.

Actually there is a [`String.split/2`](https://hexdocs.pm/elixir/String.html#split/2), which corresponds to the default `options \\ []`. Hex docs shows it properly, and as you mentioned yourself you are able to call `String.split("", ",")` 🙂

The purpose of defaults is precisely to define several arities in one signature.

sabiwara:
```elixir
iex(1)> String.split("")
[]
```

> That's a bug in Elixir.

@josevalim I agree this is inconsistent with `String.split/2/3`, but it seems intentional looking at the [tests](https://github.com/elixir-lang/elixir/blob/00f26629daa66a51efa050c51e8c5ca1ab62eeca/lib/elixir/test/elixir/string_test.exs#L30) and [code](https://github.com/elixir-lang/elixir/blob/00f26629daa66a51efa050c51e8c5ca1ab62eeca/lib/elixir/unicode/unicode.ex#L470).

Removing the `trim` seems like a breaking change, should we document the trimming behavior instead?

sabiwara:
Random thought: we could change `String.split/2/3`'s pattern to `pattern() | Regex.t() | :whitespace`, deprecate `String.split/1`, and `String.split(s)` would have to be replaced by `String.split(s, :whitespace, trim: true)`.
This could also expose `whitespace -- non_breakable` and make it work with more options?

josevalim:
Yes, agreed, this is intentional. 

When it comes to provide a better API, a general golden rule is to not have the return type changed by options. So the real issue here is that the return type depends on the option `trim: true | false`. So what we would want to do in the future is to deprecate the `trim` option and introduce `split_and_trim`, which may return an empty list.

`:whitespace` could be a choice if we want to allow whitespace to be used with other options.

But for now, I think we are good, especially because `String.split/2` does behave exactly as described in the post. :)

zorn:
> The purpose of defaults is precisely to define several arities with one signature.

Interesting. This is kind of a new concept for me. In the past, if a function accepted three arguments, even if the third one was optional, I would still always use the `/3` declaration.

But as you say, because the third argument is optional, the language infers that there is a `/2` version.

As seen trying to tab complete the function name in an `iex` session:

```
iex(5)> String.split
split/1       split/2       split/3       split_at/2    splitter/2    splitter/3
```

Thanks for the quick responses and guidance. 👍

Issue #1781, opened 2024-11-04
install.sh: OTP master fails to install without obvious error


zachallaun:
cc @wojtekmach 

Using the latest `install.sh` from https://elixir-lang.org/install.sh:

```sh
~/dev/misc
❯ sh install.sh elixir@main otp@master
downloading https://github.com/elixir-lang/elixir/releases/download/main-latest/elixir-otp-27.zip
downloading https://builds.hex.pm/builds/otp/amd64/ubuntu-22.04/master.tar.gz
unpacking elixir-otp-27.zip to /home/zacha/.elixir-install/installs/elixir/main-otp-27...
curl: (92) HTTP/2 stream 0 was not closed cleanly: PROTOCOL_ERROR (err 1)
checking OTP... 27 ok
checking Elixir... 1.18.0-dev ok

Run this (or add to your ~/.bashrc or similar file):

    export PATH=$HOME/.elixir-install/installs/otp/master/bin:$PATH
    export PATH=$HOME/.elixir-install/installs/elixir/main-otp-27/bin:$PATH

~/dev/misc 12s
❯ ls ~/.elixir-install/installs
elixir
```

Note the "HTTP/2 stream 0 was not closed cleanly" error, followed by the "ok" check that passed despite the download/install failing. 

Things succeed when downloading `otp@latest`:

```sh
~/dev/misc
❯ sh install.sh elixir@main otp@latest
downloading https://builds.hex.pm/builds/otp/amd64/ubuntu-22.04/OTP-27.1.2.tar.gz
unpacking OTP-27.1.2.tar.gz to /home/zacha/.elixir-install/installs/otp/27.1.2...
checking OTP... 27 ok
checking Elixir... 1.18.0-dev ok

Run this (or add to your ~/.bashrc or similar file):

    export PATH=$HOME/.elixir-install/installs/otp/27.1.2/bin:$PATH
    export PATH=$HOME/.elixir-install/installs/elixir/main-otp-27/bin:$PATH


~/dev/misc 33s
❯ ls ~/.elixir-install/installs
elixir  otp
```

zachallaun:
Hmm... Perhaps this was a transient error with builds.hex.pm? Because it's suddenly working.

I suppose it's worth leaving this open to track error handling when downloads fail?

wojtekmach:
[...]

PR #1782, opened 2024-11-14
Ensure install.sh exits if either of OTP or Elixir fails to install

zachallaun:
This script uses `set -e` to immediately exit the program if any commands return non-zero, but `wait` without arguments does not reliably return with the exit code of a failed subprocess. (Exactly which exit code is returned depends, I believe, on the order in which the subprocesses exit.)

By waiting on both pids explicitly, we ensure that the script exits if either OTP or Elixir fails to install. In the case of #1781, this change makes it so that the script halts after the `curl` fails to download OTP master instead of carrying on as if nothing failed.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -107,8 +107,13 @@ main() {
 
   if unzip_available; then
     install_otp &
+    pid_otp=$!
+
     install_elixir &
-    wait
+    pid_elixir=$!
+
+    wait $pid_otp
+    wait $pid_elixir
   else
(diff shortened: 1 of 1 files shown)


zachallaun:
Note that I didn't change install.bat because it does not install in parallel and so doesn't have this issue.

josevalim:
cc @wojtekmach

wojtekmach (approved):

josevalim:

PR #1787, opened 2025-01-06
Fix minor typo in index.html

ravicodelabs:
This was a minor typo fix and therefore I didn't test the change by re-building the docs. If I've missed something, please let me know and I can take a closer look.


whatyouhide:
Thanks @ravicodelabs 💟

PR #1790, opened 2025-01-19
Shorten Fedora dnf install command

leifmetcalf:
From the dnf docs:

`--repo=<repoid>, --repoid=<repoid>`
    Enable just specific repositories by an id or a glob. Can be used multiple times with accumulative effect. It is basically a shortcut for `--disablerepo="*" --enablerepo=<repoid>` and is mutually exclusive with the `--disablerepo` option.


josevalim:

PR #1797, opened 2025-03-25
Add missing deps to Gemfile

lawik:
I hit some snags setting up due to these gems not being in my ASDF-based install. If that's weird, close this, otherwise probably helps people contribute.


josevalim:

PR #1798, opened 2025-03-26
Improve sorting of cases

lawik:
- Add flagship boolean which will pin a case to the top of cases
- Sort all cases (including flagships) by date, most recent first
- Added suggested cases: Discord, Heroku and Remote


lawik:
I couldn't see a nice way to invert a `where 'flagship'` filter using liquid templates. So I went with dumb and simple.

lawik:
Change this:

Screenshot_20250326_095958


To this:

Screenshot_20250326_095104
Screenshot_20250326_095125

We can skip the change in visuals if it is a point of contention. I just found it a bit clunky and tried to tidy it up :)

lawik:
Closed pending redesign. Replaced with #1799

PR #1799, opened 2025-03-26
Improve sorting of cases

lawik:
- Add flagship boolean which will pin a case to the top of cases
- Sort all cases (including flagships) by date, most recent first
- Added suggested cases: Discord, Change.org and Remote


josevalim:

PR #1803, opened 2025-07-11
Fix Exercism icon and links

iHiD:
Fixes three things:
- Exercism is normally refered to as Exercism (not Exercism.org or Exercism.io). 
- The site moved from .io to .org a number of years back. 
- The icon URL was an old outdated one which is no longer loading.


josevalim:

PR #1824, opened 2026-05-09
Add link to mise

### LINKED ISSUE #1822: Mention mise on install page?
Hello,
I just needed to manage elixir versions on several different distros and found out about mise ( [Web](https://mise.jdx.dev), [GitHub](https://github.com/jdx/mise) ) and it's (to me) the best way to handle elixir and erlang versions, better than official package managers, better than asdf. Shouldn't it be mentioned in Install page since it's easiest way to handle elixir and erlang? 

Should be easier, faster and more secure (!) than asdf.

AlternateRT:
Also updates some of the other links, fixes some of the markdown blocks, and moves the version managers section into a level two heading

Closes #1822

### DIFF
--- a/install.markdown
+++ b/install.markdown
@@ -24,10 +24,10 @@ Install Elixir according to your operating system and tool of choice.
 
   - Using [install scripts](#install-scripts)
 
-  - Using [Homebrew](https://brew.sh/):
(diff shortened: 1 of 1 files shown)


josevalim:

PR #1834, opened 2026-07-14
Highlight home snippets using shiki

jonatanklosko:
Replaces the simple elixir tokenizer with shiki (already used by blog posts). Currently it uses github-dark theme (the blog posts use github-light), I think it fits quite nicely and definitely makes the examples more readable.

The use cases snippets and the editor examples are all highlighted at build time. For the highlighting on changes, we load a separate highlighter JS chunk, only once the user focuses the editor. The editor is mostly for small snippets, so highlighting latency should not be an issue. I also pasted a 500 LOC file and even there editing works smoothly.


josevalim:

PR #1838, opened 2026-07-16
Add cursor-pointer to stepped nav on install page

Gazler:
Everywhere else seemed to use cursor-pointer, so this felt inconsistent.


josevalim:

PR #1840, opened 2026-07-16
Use dataset API for interacting with data- attrs

jakobsen:
Please consider this PR only a suggestion.

There is no functional difference between

```js
const isRevealed = el.getAttribute("data-revealed");
```

and

```js
const isRevealed = el.dataset.revealed;
```

but to me the latter reads a little neater.

Another nice feature is that `dataset.multipleWords` in JS translates to `data-multiple-words` in HTML.

### DIFF
--- a/src/pages/learning/index.astro
+++ b/src/pages/learning/index.astro
@@ -255,7 +255,7 @@ const seo = {
     const resources = root.querySelectorAll("[data-resource]");
     const bookHeadings = root.querySelectorAll("[data-book-heading]");
     const emptyState = root.querySelector("[data-empty-state]");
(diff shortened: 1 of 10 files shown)


src/pages/learning/index.astro
```diff
@@ -277,7 +277,7 @@ const seo = {
         }
         el.classList.remove("hidden");
         if (!firstRender && !el.hasAttribute("data-revealed")) {
-          el.setAttribute("data-revealed", "true");
+        el.dataset.revealed = "true";
```
josevalim:
```suggestion
          el.dataset.revealed = "true";
```

josevalim:

PR #1841, opened 2026-07-16
Use elixir-slack.community for all Slack links

sorentwo:
The elixir-lang.slack.com address only works for existing users, whereas the elixir-slack.community site handles automatic invites for new users.


josevalim:

